PR #1, opened 2019-04-26
Add maven-publish plugin

msridhar:
With this change you can run `./gradlew publishToMavenLocal` to push a snapshot build to the local Maven repository, which makes it easy to import the checker into other projects.


msridhar:
Will open a PR against master once #3 lands

kelloggm:
We should open a new PR against master with this content.

msridhar:
Yes will do

PR #3, opened 2019-04-30
Basic draft implementation

kelloggm:
This draft implementation of the `@CalledMethods` type system works for non-fluent builder invocations like this:
```java
void foo(Builder builder) {
    builder.a();
    // type of builder is @CalledMethods({"a"})
    builder.b();
    // type of builder is @CalledMethods({"a", "b"})
}
```

There are five important parts of this pull request:
* The annotations themselves: `@CalledMethods` and `@CalledMethodsBottom`.
* The annotated type factory, which defines subtyping, LUB, and GLB on those.
* The transfer function, which performs the inference described above.
* An additional declaration annotation (`@ReturnsReceiver`) which will be a basis for handling non-fluent builders.
* Tests for the above, and some speculative tests that we'd like to be able to handle in the future, but can't yet. The former are in the `tests/basic` directory; other directories under `tests/` contain other examples we'd eventually like to be able to handle.


src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,127 @@
...
+        /**
+         * GLB in this type system is set union of the arguments of the two annotations,
+         * unless one of them is bottom, in which case the result is also bottom.
+         */
+        @Override
+        public AnnotationMirror greatestLowerBound(final AnnotationMirror a1, final AnnotationMirror a2) {
+            if (AnnotationUtils.areSame(a1, BOTTOM) || AnnotationUtils.areSame(a2, BOTTOM)) {
+                return BOTTOM;
+            }
+            Set<String> a1Val = getValueOfAnnotationWithStringArgument(a1).stream().collect(Collectors.toSet());
```
msridhar: Why use a stream here?  I'd prefer just `new LinkedHashSet<>(...)` unless there's a good reason to do otherwise.  Same for other instances of `xxx.stream().collect(Collectors.toSet())`

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderTransfer.java
```diff
@@ -0,0 +1,76 @@
...
+        if (AnnotationUtils.areSame(type, atypefactory.BOTTOM)) {
+            return result;
+        }
+
+        String methodName = node.getTarget().getMethod().getSimpleName().toString();
+        List<String> newList = new ArrayList<>();
+        newList.add(methodName);
+        List<String> currentMethods = TypesafeBuilderAnnotatedTypeFactory.getValueOfAnnotationWithStringArgument(type);
+        newList.addAll(currentMethods);
+
```
msridhar: More concise alternative (up to you on whether to use it):

```java
List<String> newList = Stream.concat(Stream.of(methodName), currentMethods.stream()).collect(Collectors.toList());
```

kelloggm:
@msridhar I've implemented the changes you suggested wrt `Stream`s.

On the topic of the Guice example, @martinschaef suggested it (along with the examples in the `tests/lombok` and `tests/cve`, both of which are definitely instances of unsafe builders). I don't quite understand it, either, but he committed it directly onto this branch, so I've included it in this PR even though it doesn't really belong.

Hopefully he and I will be able to talk on the phone today or tomorrow, and we'll document why he thinks this is an instance of the problem we're interested in. 

I think we can go ahead and merge this branch in the meantime, though, because only the tests in the `basic` directory are actually being run by the testing infrastructure.

martinschaef:
The Guice example is probably not applicable by this implementation but could be handled by a similar type system. A Guice 'module' has a number of @Provides methods that bind the return type of the method to a particular object for all uses of this type. E.g., in

@Provides
private Integer providesFive() { return 5; }
@Provides
private String providesBla(final Integer x) { ... }

The value of the integer x will always be 5. 

The issue is that for every parameter that you use in an @Provides-annotated method, there needs to be another @Provides-annotated method in the same module or an 'installed' module that returns this type. If not, you get a runtime exception. In a way this is a type-state problem as well, but there is no control-flow that we need to track.

msridhar:
@martinschaef time to switch to Dagger 😛 The Guice example is interesting and it'd be cool to catch it with a type system, but to me it seems unrelated to tracking called methods on builders.  In fact, I think it requires something like a global analysis (to know all the available modules) which might put it a bit outside the scope of Checker-Framework-style checkers.

martinschaef:
Let's talk about it more later. Instead of tracking invocations of builder methods you would need to track declarations of provider methods ... so its probably even more of a type problem because you don't need flow. However, it is a very different type system.

PR #4, opened 2019-04-30
Add support for fluent builders in the presence of @ReturnsReceiver annotations

kelloggm:
This pull request adds support for fluent builders to the `wip` branch.

In the presence of an `@ReturnsReceiver` annotation, the annotated type factory now infers `@CalledMethods` annotations for chained builders.

This implementation is unsound unless the `@ReturnsReceiver` annotation is checked. In this version of the code, it is trusted; we plan to implement a check soon.

### DIFF
--- a/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
+++ b/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
@@ -4,7 +4,7 @@ import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
 import org.checkerframework.checker.builder.qual.CalledMethods;
 import org.checkerframework.checker.builder.qual.CalledMethodsBottom;
-import org.checkerframework.checker.index.IndexUtil;
+import org.checkerframework.checker.builder.qual.ReturnsReceiver;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
 import org.checkerframework.framework.type.AnnotatedTypeFactory;
@@ -15,8 +15,10 @@ import org.checkerframework.framework.type.treeannotator.TreeAnnotator;
 import org.checkerframework.framework.util.MultiGraphQualifierHierarchy;
 import org.checkerframework.javacutil.AnnotationBuilder;
 import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.TreeUtils;
 
 import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.Element;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


kelloggm:
The difference is fundamental to the difference between the annotated type factory and the transfer function's store. Each is useful for different kinds of refinements:
* the annotated type factory produces a type for every expression in the program. However, it is not suitable for flow-sensitive refinement or for refining the type of objects based on their context (though the latter is possible if you abuse the system a bit)
* the transfer function permits flow-sensitive (or other kinds of context-sensitive) refinement, but can only express types for certain kinds of expressions (those with an extremely-poorly-named `Receiver` in the internal representation used by the CFG).

Those distinctions are why the two different refinements are implemented in different places.
* the original refinement (from the `wip` branch) modifies the type of a receiver based on the context: it operates on a CFG node, but modifies the receiver. This works as long as the store contains the expression, which will only be true if the receiver is an identifier or implicit - the store doesn't hold `new` expressions or chains of method calls
* the refinement implemented by this PR changes the type of a method invocation. The type of a method invocation is its return type, so when we know that the receiver *is* the return type. Since we're interested in fluent builders (which are chains of method calls - so they wouldn't be in the store used by the transfer function), the type factory makes sense as the right place to do the refinement.

msridhar:
Not sure I fully understand, @kelloggm.  The receiver type for a method may be dependent on the flow-sensitive refinement, so it seems like in this case the annotated type factory is potentially using results of the refinement, even if it doesn't do so directly.  I don't have a concrete example, but what if the relevant code occurs in a loop such that it takes multiple iterations for the receiver type to reach a fixed point.  Will this approach still work in such a case?

msridhar:
I wrote this loopy test to see what would happen:
```java
    static void fluentLoop() {
        SimpleFluentInference s = new SimpleFluentInference().a();
        int i = 10;
        while (i > 0) {
            s.b().build();
            i--;
            s = new SimpleFluentInference();
        }
    }
```
An error is reported at the `build()` call as desired.  So the current implementation seems to handle loops (unless this was a fluke of dataflow iteration order), which means the type factory is likely getting invoked multiple times to get the type of `s.b()`.  I'd still like to understand better when exactly the type factory gets consulted as to the type of an expression and its relationship with flow-sensitive type refinement; is there a doc somewhere I can look at?

kelloggm:
Sorry for being slow to respond to this.

The best documentation I know of is in this chapter of the CF manual: https://checkerframework.org/manual/#creating-a-checker, which describes the various parts and what they should do. Most of what I wrote above I learned just by writing a bunch of typecheckers, though - you develop an intuition for what should be done where, and I was trying to explain it (but I might have done so somewhat poorly).

In the loop example you wrote above, the type of `s` at the start of the loop is the least upper bound of the type coming in along each control flow edge. There are two: the edge coming from the start of the loop (on which the type of `s` is `@CalledMethods({"a"})` and the one from the end of the loop (on which the type of `s` is `@CalledMethods({})`. The LUB of these is `@CalledMethods({})`, since I defined the LUB of called methods annotations as intersection. 

To compute that the type of `s` at the end of the loop is `@CalledMethods({})`, the framework does have to evaluate `s.b()` more than once. The framework will go around the loop over and over so long as the types change on each iteration (though I think for practical reasons there is a "go to top after X iterations" rule).

Hopefully that helps. If not, I'm happy to set up a call to try explain what's happening.

kelloggm:
I'm also going to close this PR and re-open it against master.

PR #8, opened 2019-05-02
Implementation of @CalledMethodsPredicate

kelloggm:
This pull request adds support for a new annotation in the `@CalledMethods` type system: `@CalledMethodsPredicate`. This annotation exists to generically extend the possibly least upper bounds that are expressible in this type system by permitting disjunctions.

The motivation for this addition to the type system is that there are builders (and other kinds of situations) where we want to permit a few different _combinations_ of called methods. Take, for example, this example in `tests/cve/Cve.java`, which is an AWS API that is vulnerable to a so-called "AMI sniping" attack (which permits malicious code injection) if one of the logical parameters is specified without another. However, another safe set of calls uses a totally different parameter, and ignores the other two calls. The stub file with the (full) specification for this API is in the `stubs/DescribeImages.astub` file (the reason for the proliferation of `||`s in the spec is that these AWS APIs have two methods to supply each parameter: one fluent, and one side-effectful).

I implemented a simple predicate language using the [Javaluator](http://javaluator.sourceforge.net/en/home/) framework, which seemed like a good choice because I didn't want to re-implement a recursive descent parser from scratch.

The primary changes in this pull request are:
- The new annotation `@CalledMethodsPredicate`, and a new unified top annotation so that `@CalledMethodsPredicate` and `@CalledMethods` are siblings in the type hierarchy and are unassignable except via the mechanism in the new subtyping relationship
- An evaluator, built on top of Javaluator, which determines whether a given predicate is true or false given a set of methods that have been called on an object
- Modifications to the subtyping, glb, and lub algorithms to support the new annotation
- Basic tests for the new annotation and some simple predicates
- A visitor for the type system that rejects (some) mis-formed predicates with a special error message
- A test that enforces that predicates with unbalanced parentheses are rejected
- A new test set that expands on Martin Schaef's CVE test by including both correct and incorrect calls, and both fluent and non-fluent examples

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,8 +15,13 @@ dependencies {
     // This dependency is found on compile classpath of this component and consumers.
     implementation 'org.checkerframework:checker:2.8.1'
 
+    // For the @CalledMethodsPredicate evaluation
+    implementation 'com.fathzer:javaluator:3.0.2'
+    
     // Use JUnit test framework
     testImplementation 'junit:junit:4.12'
+    
+    // CF testing infrastructure
     testCompile 'org.checkerframework:testlib:2.5.4'
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,9 @@
 plugins {
     // Apply the java plugin to add support for Java
     id 'java'
+
+    // For bringing in AWS dependencies
+    id 'io.spring.dependency-management' version '1.0.7.RELEASE'
 }
 
 repositories {
@@ -17,12 +20,21 @@ dependencies {
 
     // For the @CalledMethodsPredicate evaluation
     implementation 'com.fathzer:javaluator:3.0.2'
-    
+
     // Use JUnit test framework
     testImplementation 'junit:junit:4.12'
-    
+
     // CF testing infrastructure
     testCompile 'org.checkerframework:testlib:2.5.4'
+
+    // For CVE tests
+    testCompile 'com.amazonaws:aws-java-sdk-ec2'
+}
+
+dependencyManagement {
+    imports {
+        mavenBom 'com.amazonaws:aws-java-sdk-bom:1.11.228'
+    }
 }
 
 tasks.withType(JavaCompile).all {
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,10 @@ plugins {
     // Apply the java plugin to add support for Java
     id 'java'
 
-    // For bringing in AWS dependencies
+    // For bringing in AWS dependencies. We're using the "legacy" version
+    // listed here: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-project-gradle.html,
+    // because the build fails when we use the Gradle 5+ way, saying that it can't find ec2.
+    // I believe that this is because we need the AWS SDK during testing, but not during the build.
     id 'io.spring.dependency-management' version '1.0.7.RELEASE'
 
     id 'maven-publish'
--- a/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
+++ b/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
@@ -201,8 +201,7 @@ public class TypesafeBuilderAnnotatedTypeFactory extends BaseAnnotatedTypeFactor
                 // superAnno is a CMP annotation, so we need to evaluate the predicate
                 String predicate = AnnotationUtils.getElementValue(superAnno, "value", String.class, false);
                 CalledMethodsPredicateEvaluator evaluator = new CalledMethodsPredicateEvaluator(subVal);
-                String evalTree = evaluator.evaluate(predicate, new ArrayList<String>());
-                String result = evalTree.substring(evalTree.lastIndexOf('=') + 1);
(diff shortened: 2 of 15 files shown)


src/main/java/org/checkerframework/checker/builder/qual/CalledMethodsPredicate.java
```diff
@@ -0,0 +1,34 @@
...
+ * permit methods to be annotated to require a disjunction
+ * of method calls. So, for instance, if it was acceptable
+ * to call method a() or method b() before calling method c(),
+ * then c()'s receiver should have an @CalledMethodsPredicate
+ * annotation with the argument "a || b"
+ *
+ * The argument is a string. The string must be constructed
+ * from the following grammar:
+ *
+ * S → method name | (S) | S && S | S || S
```
martinschaef: If you have `||` and `&&` but not negation, it essentially amounts to enumerating all legal combinations of builder methods. Are you planning to have negation as well? Otherwise, it would be more intuitive to just have lists of lists.

martinschaef: I think, to not overcomplicate things, you would be better of if you just allow folks to enumerate legal combinations. Otherwise, we can rathole on things that look cool but have no real value for the paper (e.g., enabling things like `if you use X you must not use Y`)

kelloggm: The spec for the EC2 API is going to get really nasty if we have to enumerate all valid combinations, so I don't think I agree with you on this.

I suspect there will be even more complex requirements than that that we will want to specify, and the predicates make that somewhat more reasonable.

I have no intention to support negation. The predicates are just syntactic sugar for a list of lists.

build.gradle
```diff
@@ -1,6 +1,9 @@
 plugins {
     // Apply the java plugin to add support for Java
     id 'java'
+
+    // For bringing in AWS dependencies
```
msridhar: Why do we need a Gradle plugin to bring in dependencies?  Maybe a bit more explanation in the comment?

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -169,9 +184,30 @@ public boolean isSubtype(final AnnotationMirror subAnno, final AnnotationMirror
...
+
+            if (AnnotationUtils.areSameByClass(superAnno, CalledMethodsPredicate.class)) {
+                // superAnno is a CMP annotation, so we need to evaluate the predicate
+                String predicate = AnnotationUtils.getElementValue(superAnno, "value", String.class, false);
+                CalledMethodsPredicateEvaluator evaluator = new CalledMethodsPredicateEvaluator(subVal);
+                String result = evaluator.evaluate(predicate);
+                return Boolean.parseBoolean(result);
+            } else {
+                // superAnno is a CM annotation, so compare the sets
+                Set<String> superVal = new LinkedHashSet<>(getValueOfAnnotationWithStringArgument(superAnno));
```
msridhar: Is there a reason for `superVal`?  Doesn't `containsAll` take a `Collection` as an argument?

src/main/java/org/checkerframework/checker/builder/qual/CalledMethodsPredicate.java
msridhar: Given the spec written in `DescribeImages.astub` I think it's worthwhile to support `&&` and `||` as opposed to just a list of lists

src/main/java/org/checkerframework/checker/builder/qual/CalledMethodsPredicate.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * That is, the permitted elements are method names, parentheses,
+ * and the strings "&&" and "||". "&&" has higher precedence
+ * than "||", following standard Java operator semantics.
```
msridhar: Have you thought about handling of method overloading?  I wouldn't be surprised if it comes up at some point.  Doesn't need to block this PR but probably we should have an open issue on it (as it will force us to go beyond just method names in these annotations)

src/main/java/org/checkerframework/checker/builder/qual/CalledMethodsPredicate.java
martinschaef: Ah, I see what you are saying. So it's more like a regular expression than actual logic. Fine by me, I guess.

build.gradle
kelloggm: I followed the instructions here: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-project-gradle.html. The first method (with the change to the `settings.gradle` file) doesn't work - `testCompile` can't find the dependencies it needs. The "legacy" approach, however, does work.

kelloggm: I've added some text in a comment there that explains it.

src/main/java/org/checkerframework/checker/builder/qual/CalledMethodsPredicate.java
kelloggm: I had not thought about method overloading. I agree that there's no need to block this PR, but I'll open an issue and think about how to address it. Nice catch.

PR #12, opened 2019-05-15
Add AutoValue builder test

msridhar:
Adds a test for AutoValue Builder code.


src/test/java/tests/BasicTest.java
```diff
@@ -1,3 +1,5 @@
+package tests;
```
kelloggm: I don't think this *really* matters, but there's certainly no reason to add it in this PR.

PR #13, opened 2019-05-15
Enforce style

kelloggm:
Via the [spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-to-java-source-google-java-format) plugin, using Google's Java format (which I am a fan of, and which is used by the Checker Framework).

I am merging this change without approval, assuming that it builds on Travis, because it is non-functional.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,8 @@ plugins {
     id 'io.spring.dependency-management' version '1.0.7.RELEASE'
 
     id 'maven-publish'
+
+    id "com.diffplug.gradle.spotless" version "3.23.0"
 }
 
 repositories {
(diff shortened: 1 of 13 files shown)


mernst:
The only concern with this is that out of the box, google-java-format produces bad formatting for type annotations.  I use https://github.com/plume-lib/run-google-java-format instead, but it's a lot clunkier to set up than spotless.

kelloggm:
At least for now there are no type annotations in this code, so I preferred the simpler option (and gradle believes that the test cases, which do contain plenty of annotations, are test data, not a source set, so spotless isn't applied).

PR #15, opened 2019-05-23
[WIP] Built-in support for AutoValue Builders

msridhar:
This PR contains initial support for AutoValue Builders in the typesafe builder checker.  In particular:

* `@CalledMethod` annotations for `build()` methods in builders are added automatically
* setter methods in builders are treated as having `@ReturnsReceiver` annotations

I am not sure about much of the structure of this code.  Much of it is based on copy-pasting and modifying @kelloggm's preliminary Lombok support code.  It should be carefully reviewed before merging, and may require significant restructuring.  I will add some inline comments with further thoughts / concerns.


src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -107,115 +117,85 @@ public Void visitMethodInvocation(
...
-     * GLB in this type system is set union of the arguments of the two annotations, unless one of
-     * them is bottom, in which case the result is also bottom.
-     */
-    @Override
-    public AnnotationMirror greatestLowerBound(
-        final AnnotationMirror a1, final AnnotationMirror a2) {
-      if (AnnotationUtils.areSame(a1, BOTTOM) || AnnotationUtils.areSame(a2, BOTTOM)) {
-        return BOTTOM;
+    private boolean isAutoValueBuilderSetter(Element element) {
+      MethodTree methodTree = (MethodTree) declarationFromElement(element);
```
msridhar: I could not figure out how to insert declaration annotations in a type factory.  And making `@ReturnsReceiver` a type use annotation seemed wrong.  So I added inline logic to treat methods as having `@ReturnsReceiver` when they are setters in AutoValue builders.  We need to make this configurable eventually.

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -107,115 +117,85 @@ public Void visitMethodInvocation(
...
     }
 
-    /** isSubtype in this type system is subset */
     @Override
-    public boolean isSubtype(final AnnotationMirror subAnno, final AnnotationMirror superAnno) {
-      if (AnnotationUtils.areSame(subAnno, BOTTOM)) {
-        return true;
-      } else if (AnnotationUtils.areSame(superAnno, BOTTOM)) {
-        return false;
+    public Void visitExecutable(AnnotatedTypeMirror.AnnotatedExecutableType t, Void p) {
```
msridhar: This injects the `@CalledMethods` annotation computed by the `AutoValueBuilderAnnotatedTypeFactory`

src/main/java/org/checkerframework/checker/builder/autovalue/AutoValueBuilderAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,162 @@
...
+import org.checkerframework.framework.type.typeannotator.ListTypeAnnotator;
+import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
+import org.checkerframework.framework.util.MultiGraphQualifierHierarchy;
+import org.checkerframework.framework.util.MultiGraphQualifierHierarchy.MultiGraphFactory;
+import org.checkerframework.javacutil.AnnotationBuilder;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.TreeUtils;
+
+/** Responsible for placing appropriate annotations on Lombok builders. */
+public class AutoValueBuilderAnnotatedTypeFactory extends BaseAnnotatedTypeFactory
```
msridhar: This is the part I am most conflicted about.  I understand we want to have built-in framework support be configurable, and doing it via separate type factories kind of makes sense.  But in terms of code structure it is a mess, particularly because it seems the `TypesafeBuilderAnnotatedTypeFactory` cannot get these inserted annotations "for free" and has to again re-insert them.  Maybe the code is just wrong?  I think we should consider alternatives for configuring framework support, such as configurable options within the type safe builder checker.  That won't let users plug in their own framework support, but properly supporting that would be a bunch more work too (I think).

kelloggm:
@msridhar I agree that this code ended up looking like spaghetti. I think the design I had in mind when I was building the Lombok version of this code made sense when we expected at least two frameworks to be supported in this manner (i.e. Lombok and AutoValue), but now I don't think that will end up being the case (because of the issues we ran into with Lombok).

I propose eliminating the separate AutoValueChecker and integrating the code in it into the main TypesafeBuilderChecker. I think that will dramatically reduce the messiness of this code, and we can eliminate the refactorings I did to pull parts of the `TypesafeBuilderAnnotatedTypeFactory` out so that they could be re-used.

In terms of implementation, I think what we should do is start on a fresh branch based on `master` and copy only the core pieces of code from this PR into that branch:
* the `visitExecutable` code in `AutoValueTypeAnnotator`
* the `hasReturnsReceiver` method and its callsites in `TypesafeBuilderAnnotatedTypeFactory`
* the tests

Is there anything else that we actually need from this pull request?

msridhar:
@kelloggm that sounds reasonable (fresh branch from master).  I'll leave this PR open until that is ready and merged.  I think you covered all the things that are needed.

msridhar:
Closing in favor of #17

PR #16, opened 2019-06-18
Integrate returns receiver checker

kelloggm:
This branch implements a locally-correct integration between the returns receiver checker (https://github.com/msridhar/returnsrecv-checker) and this typechecker.

I do not expect the travis build to pass, because that checker is a dependency, but isn't (yet) available publicly. We shouldn't merge this until it is.

I had to disable running the checker on the code AutoValue generates to get those tests to pass. That's a pretty lame solution to the problem of needing to automatically add `@This` annotations to the generated code, so it eventually needs to be fixed. Otherwise the changes are straightforward.


build.gradle
```diff
@@ -27,6 +28,9 @@ dependencies {
 
     // For the @CalledMethodsPredicate evaluation
     implementation 'org.springframework:spring-expression:5.1.7.RELEASE'
+    
+    // For returns receiver checking
+    implementation 'org.checkerframework:returnsrcvr-checker:0.1-SNAPSHOT'
```
msridhar: Maybe we can use https://jitpack.io/ to pull in this dependence without having to upload to Maven Central?  This would only be temporary.

kelloggm:
Great idea!

PR #17, opened 2019-06-20
Implement @CalledMethods insertion for AutoValue Builders

msridhar:
Support for inserting `@This` annotations for AutoValue Builders was added to the returns receiver checker in msridhar/returnsrecv-checker#1.  Building on that, this PR implements `@CalledMethods` insertion for AutoValue Builders.  With this change, the checker is able to check AutoValue Builder usage for correctness with no manual annotations required; see updates to the `Animal.java` test case.

### DIFF
--- a/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
+++ b/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
@@ -1,9 +1,19 @@
 package org.checkerframework.checker.builder;
 
 import com.google.auto.value.AutoValue;
-import com.sun.source.tree.*;
+import com.sun.source.tree.AnnotationTree;
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
 import java.lang.annotation.Annotation;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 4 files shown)


src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -1,28 +1,29 @@
...
 
-import com.sun.source.tree.ExpressionTree;
-import com.sun.source.tree.MethodInvocationTree;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
+import com.google.auto.value.AutoValue;
+import com.sun.source.tree.*;
```
kelloggm: We stylistically prefer no wildcard import in Checker Framework code, so this will be rejected if we try to move it into the main repository. I acknowledge I don't have a lint rule turned on for it right now in this repo, though, so if you want to leave it it's fine.

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -119,6 +126,100 @@ public Void visitMethodInvocation(
...
+      }
+
+      return super.visitExecutable(t, p);
+    }
+
+    /**
+     * computes the required properties of an @AutoValue class
+     *
+     * @param autoValueClass
+     * @return
```
kelloggm: I'd either fill out the Javadoc or remove it

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -119,6 +126,100 @@ public Void visitMethodInvocation(
...
+     * creates a @CalledMethods annotation for the given property names, converting the names to the
+     * corresponding setter method name in the Builder
+     *
+     * @param propertyNames the property names
+     * @return the @CalledMethods annotation
+     */
+    public AnnotationMirror createCalledMethodsForProperties(final List<String> propertyNames) {
+      String[] calledMethodNames =
+          propertyNames.stream()
+              .map((prop) -> "set" + prop.substring(0, 1).toUpperCase() + prop.substring(1))
```
kelloggm: Does AutoValue permit the user to change the names of generated methods? I know Lombok does, and it's a huge pain to deal with.

If so, I'm not sure the assumptions about the names of methods here will hold. I'm not sure we can really do any better, though, so maybe it's okay.

msridhar: I am actually not 100% sure what AutoValue allows in builders.  I'd have to dig into the implementation to check.  But in my experience I've only seen the `setXXX()` syntax.  We will probably have to handle a bunch of additional corner cases when we test on real code, so I'm not going to pre-emptively investigate.

kelloggm: That seems reasonable.

PR #18, opened 2019-06-27
Add lombok support

kelloggm:
This code adds support for automatically adding `@CalledMethods` automatically to the `build` method in Lombok-generated builders.

It depends on https://github.com/msridhar/returnsrecv-checker/pull/2; the tests will **not** pass until that PR is merged.

Much of the logic is shared with AutoValue builders; I generalized it to include Lombok builders.

I also changed the Lombok tests so that they use the delombok'd version - this code does not work on non-delombok'd code, due to general issues with the Checker Framework.


kelloggm:
@ranmanli I would also appreciate a review from you

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -158,38 +165,100 @@ public Void visitExecutable(AnnotatedTypeMirror.AnnotatedExecutableType t, Void
...
+     */
+    private List<String> getLombokRequiredProperties(final ClassTree lombokClass) {
+      List<String> requiredPropertyNames = new ArrayList<>();
+      List<String> defaultedPropertyNames = new ArrayList<>();
+      for (Tree member : lombokClass.getMembers()) {
+        if (member.getKind() == Tree.Kind.VARIABLE) {
+          VariableTree fieldTree = (VariableTree) member;
+          for (AnnotationTree atree : fieldTree.getModifiers().getAnnotations()) {
+            AnnotationMirror anm = TreeUtils.annotationFromAnnotationTree(atree);
+            if (AnnotationUtils.areSameByClass(anm, lombok.NonNull.class)) {
```
msridhar: Are we sure that Lombok checks for `@lombok.NonNull` and not any annotation named `@NonNull`?

kelloggm: Good idea. I've changed the check so that it's consistent with the list of annotations that Lombok uses.

PR #19, opened 2019-07-06
Add manifest file for processor auto-discovery

msridhar:
This will allow us to run the checker without adding a `-processor` flag, by just adding the jar to the processor path, leveraging processor auto-discovery.  This is important for running on projects that already use processor auto-discovery.

### DIFF
--- a/src/main/resources/META-INF/services/javax.annotation.processing.Processor
+++ b/src/main/resources/META-INF/services/javax.annotation.processing.Processor
@@ -1,0 +1,1 @@
+org.checkerframework.checker.builder.TypesafeBuilderChecker


msridhar:
A small downside here is that if in the future some dependent checker depends on this checker, this manifest file will make the checker run even if that is not desired.  I think that is a corner case we don't need to worry about for now, but we wouldn't want to add a similar manifest file to the returns receiver checker for this reason.

kelloggm:
I think this is fine, at least as a temporary solution.

The Checker Framework's standard practice is to not provide a manifest file, because the framework has many checkers. Since this repository only contains one checker, I think including a manifest file is okay - if a user has this on their classpath, there must be a reason.

PR #21, opened 2019-07-08
Fixes for AutoValue support

msridhar:
1. Don't treat non-`abstract` methods as corresponding to required properties
2. Handle more sophisticated mappings between AutoValue method names and the corresponding Builder method names

### DIFF
--- a/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
+++ b/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
@@ -480,9 +480,11 @@ public class TypesafeBuilderAnnotatedTypeFactory extends BaseAnnotatedTypeFactor
   private static String autoValuePropToBuilderSetterName(
       String prop, Set<String> allBuilderMethodNames) {
     // property name may be prefixed JavaBean style with 'get' or 'is'; strip that part if present
-    if (prop.startsWith("get") && Character.isUpperCase(prop.charAt(3))) {
+    if (prop.startsWith("get") && prop.length() > 3 && Character.isUpperCase(prop.charAt(3))) {
       prop = Introspector.decapitalize(prop.substring(3));
-    } else if (prop.startsWith("is") && Character.isUpperCase(prop.charAt(2))) {
+    } else if (prop.startsWith("is")
+        && prop.length() > 2
+        && Character.isUpperCase(prop.charAt(2))) {
       prop = Introspector.decapitalize(prop.substring(2));
     }
     // setter name may be the property name itself, or prefixed by 'set'
--- a/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
+++ b/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
@@ -8,6 +8,7 @@ import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
+import java.beans.Introspector;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -15,6 +16,7 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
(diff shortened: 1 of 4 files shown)


msridhar:
I can't get the tests to pass here, and I am getting weird errors:

```
7 unexpected diagnostics were found:
AutoValue_GetAnimal.java:95: error: (override.receiver.invalid)
AutoValue_GetAnimal.java:103: error: (override.receiver.invalid)
AutoValue_GetAnimal.java:108: error: (override.receiver.invalid)
AutoValue_GetAnimal.java:113: error: (override.receiver.invalid)
AutoValue_Animal.java:81: error: (override.receiver.invalid)
AutoValue_Animal.java:89: error: (override.receiver.invalid)
AutoValue_Animal.java:94: error: (override.receiver.invalid)
```

The `override.receiver.invalid` errors come from the returns receiver checker, e.g.:
```
/Users/msridhar/git-repos/typesafe-builder-checker/tests/build/testclasses/AutoValue_GetAnimal.java:95: error: [override.receiver.invalid] Incompatible receiver type
    GetAnimal.Builder setName(String name) {
                      ^
  Method
    @This Builder setName(@This Builder this, @MaybeThis String p0) in AutoValue_GetAnimal.Builder
  cannot override
    @MaybeThis Builder setName(@MaybeThis Builder this, @MaybeThis String p0) in GetAnimal.Builder
  found   : @This Builder
  required: @MaybeThis Builder
```
The strange thing is if I delete the new `GetAnimal.java` test file, all the `override.receiver.invalid` errors disappear, including the ones in `AutoValue_Animal.java`, which is an unrelated class!  Also I added a minor variant of `GetAnimal.java` as a test for the returns receiver checker itself and got no `override.receiver.invalid` errors. 

@kelloggm any idea what is going on here?

msridhar:
Also if I remove `Animal.java` then the tests pass.  It's only when both `Animal.java` and `GetAnimal.java` are present that the `override.receiver.invalid` errors appear.

kelloggm:
I spent some time looking into this. I don't have a solution yet, but I've positively concluded that the problem is in the returns receiver checker, and I have a moderately well-supported hypothesis for what is wrong.

> Also I added a minor variant of GetAnimal.java as a test for the returns receiver checker itself and got no override.receiver.invalid errors.

I was surprised by this claim, and decided to test it for myself. I moved the `GetAnimal.java` file to `returnsrecv-checker/tests/autovalue/`, and got four `override.receiver.invalid` errors:
```
java.lang.AssertionError: The test run was not expected to issue errors/warnings, but it did.
0 out of 0 expected diagnostics were found.
4 unexpected diagnostics were found:
AutoValue_GetAnimal.java:95: error: (override.receiver.invalid)
AutoValue_GetAnimal.java:103: error: (override.receiver.invalid)
AutoValue_GetAnimal.java:108: error: (override.receiver.invalid)
AutoValue_GetAnimal.java:113: error: (override.receiver.invalid)
```
These look like the same errors you were getting from the typesafe builder checker, so that's good. I noticed that the other test (`Animal.java`) in the RR checker omits the *uses* of the builder, so I tried removing those from my copy of `GetAnimal` as well. To my surprise, the errors disappeared. I assume that this is some quirk of AutoValue; perhaps the classes aren't generated if they aren't being used?

After re-adding that code, I tried to trace the source of the error. Reading the long error message, I noticed that the `@This` annotation appears to have been placed correctly in the generated code, but **not in the original code**:
```
    @This Builder setName(@This Builder this, @MaybeThis String p0) in AutoValue_GetAnimal.Builder
  cannot override
    @MaybeThis Builder setName(@MaybeThis Builder this, @MaybeThis String p0) in GetAnimal.Builder
```
To test whether this was the cause of the failure, I manually inserted `@This` annotations in the original code, and the test began to pass. I therefore think that there's something off about the logic for inserting `@This` annotations into builders. We didn't catch it because the `Animal` test in the RR checker apparently isn't actually generating code, which we should fix. I don't know the AutoValue code well enough to diagnose what's wrong with it (though I might have broken it when I added Lombok support). Can you take a look @msridhar and see if there's something wrong there?

msridhar:
Thanks for digging! So weird that keeping the uses makes a difference. I
will dig further and figure this out. Thanks, Manu

msridhar:
When rebased on #24 and with the latest returns-receiver fixes, the tests pass locally for me now for this PR.  Will update the PR once #24 lands

src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -465,6 +477,27 @@ private static boolean hasAnnotation(
         .anyMatch(anm -> AnnotationUtils.areSameByClass(anm, annotClass));
   }
 
+  private static String autoValuePropToBuilderSetterName(
+      String prop, Set<String> allBuilderMethodNames) {
+    // property name may be prefixed JavaBean style with 'get' or 'is'; strip that part if present
+    if (prop.startsWith("get") && Character.isUpperCase(prop.charAt(3))) {
```
kelloggm: Is it possible for the entire name of the method to be `get()`? For example, what if I named the property "get". Would it be possible for me to direct AutoValue to produce a method that doesn't have a prefix and then cause this code to throw a `StringOutOfBoundsException`

msridhar: Good catch!  I added a fix and a test

PR #24, opened 2019-07-09
Update to Checker Framework 2.9.0

msridhar:
Fix meta-annotations for compatibility.  Note that the tests will not pass until the returns-receiver checker is also updated; see msridhar/returnsrecv-checker#6.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,8 +32,8 @@ dependencies {
     
     // For returns receiver checking. The first uses the master branch; the second
     // uses your local copy.
-    implementation 'com.github.msridhar:returnsrecv-checker:master-SNAPSHOT'
-    // implementation 'org.checkerframework:returnsrcvr-checker:0.1-SNAPSHOT'
+    // implementation 'com.github.msridhar:returnsrecv-checker:master-SNAPSHOT'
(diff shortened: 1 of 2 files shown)


msridhar:
@kelloggm this one is ready to land now I think

kelloggm:
For a standard type system like `@CalledMethods`, there's no need to explicitly specify that the bottom type is the default for `java.lang.Void` and `null` - those are the defaults. I've made the change, so this can be merged when tests pass.

build.gradle
```diff
@@ -32,8 +32,8 @@ dependencies {
     
     // For returns receiver checking. The first uses the master branch; the second
     // uses your local copy.
-    implementation 'com.github.msridhar:returnsrecv-checker:master-SNAPSHOT'
-    // implementation 'org.checkerframework:returnsrcvr-checker:0.1-SNAPSHOT'
+    // implementation 'com.github.msridhar:returnsrecv-checker:master-SNAPSHOT'
+    implementation 'org.checkerframework:returnsrcvr-checker:0.1-SNAPSHOT'
```
msridhar: @kelloggm don't think we want this change

kelloggm: Yep, removed it. Sorry.

PR #26, opened 2019-07-09
Refactor code to not require trees

msridhar:
Similar to msridhar/returnsrecv-checker#8.  With multiple rounds of annotation processing, we cannot rely on ASTs for all relevant code being present in all rounds.  This issue was exposed on a larger benchmark.  Fixing the issue exposes a bug in our AutoValue `toBuilder` support (we were not injecting annotations for the generated `toBuilder` implementation).  I have temporarily commented out that test so it can be fixed in a separate follow-up PR.


msridhar:
FYI, I have a follow-up change to fix the `toBuilder` support.

PR #27, opened 2019-07-09
Handle weird cases where 'is' and 'get' are part of property name

msridhar:
In some cases AutoValue decides to not strip 'is' and 'get' from property names, so they become part of the name.  Rather than figuring out the exact logic behind this scenario, we just try a few possibilities for setter names in the Builder.


msridhar:
Actually just realized this logic isn't quite right.  Going to try again

PR #33, opened 2019-07-15
Instructions for using the checker with Lombok

kelloggm:
Adds another README that details how to configure a project that uses Lombok to use this checker to detect errors.


mernst:
@kelloggm Can you review my changes and then squash-and-merge?
After that, I want to do some more playtesting before publicizing this.

kelloggm:
@mernst, your changes look great. Thank you!

I had to fix one minor error in `@CalledMethodsPredicate` syntax in the README, but that is just evidence that the existing README was not specific enough.

PR #35, opened 2019-07-16
Fix handling of Guava Immutable types

msridhar:
AutoValue has special handling of properties with a Guava Immutable type.  For such a property `foo`, if the AutoValue Builder has a method `fooBuilder()`, then the Builder automatically populates the property with an empty collection if `fooBuilder()` is never called.  So in such cases, don't treat `foo` as a required property.


msridhar:
I think the difference is clear. The difference is in the methods inside the nested Builder type and how that affects AutoValue’s behavior

PR #38, opened 2019-07-17
Add test that lombok builders work even when defined in bytecode

kelloggm:
I'm happy to report that this does, in fact, work as expected.

This PR adds a test that enforces that it continues to. The test:
* builds a simple Lombok'd builder jar
* adds that jar to the `testCompile` classpath so that it is used when compiling test code, but the sources are unavailable
* enforces that an error is issued when the builder defined in the jar is misused

Some non-trivial changes in the build script were necessary to make this happen.


msridhar:
Would it be cleaner to do the following?
* Introduce multiple sub-modules, one for the checker and one for the lombok test input code
* Build the Lombok sub-module as one would usually build with Gradle + Lombok
* Add a `testCompile` dependence from the checker module to the Lombok module

If it works, that would get rid of the need to explicitly write a shell script to run Lombok, etc.  We do this in NullAway, see these links:

* A [test-java-lib](https://github.com/uber/NullAway/tree/master/test-java-lib) module that is used as input for the NullAway tests
* [testCompile dependence](https://github.com/uber/NullAway/blob/master/nullaway/build.gradle#L56) from the main NullAway module on the test-java-lib module

With the above, NullAway sees only the class files for test-java-lib, not the source files.  With this approach you also need a settings.gradle at the top level including all the modules, like [this one](https://github.com/uber/NullAway/blob/master/settings.gradle) for NullAway

kelloggm:
That would be cleaner. I'll implement it.

kelloggm:
Update: this PR is now ready for review again. I followed @msridhar's suggestions on how to organize the project, and got rid of the gross shell script, which is great. I also realized that I never committed the lombok test runner, which was making my tests kind of useless on computers other than mine. So I've added that to the PR, too!

PR #40, opened 2019-07-18
Remove runtime dependency on Lombok

### LINKED ISSUE #39: Remove checker's runtime dependence on lombok jar
Unfortunately, Lombok does not distribute an annotations jar.  Since we are doing some runs with annotation processor auto-discovery, this means that when we run our checker, Lombok also gets run, which slows down the checker and could have unanticipated consequences.  It's not ideal, but we should probably use a strings to identify Lombok annotations to get rid of this dependence.

kelloggm:
By replacing the check for a specific annotation by classname to a check via a String.

Fixes #39

### DIFF
--- a/typesafe-builder-checker/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
+++ b/typesafe-builder-checker/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
@@ -247,7 +247,7 @@ public class TypesafeBuilderAnnotatedTypeFactory extends BaseAnnotatedTypeFactor
         assert hasAnnotation(nextEnclosingElement, AutoValue.class)
             : "class " + nextEnclosingElement.getSimpleName() + " is missing @AutoValue annotation";
 
-      } else if (hasAnnotation(enclosingElement, lombok.Generated.class)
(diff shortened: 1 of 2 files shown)


typesafe-builder-checker/src/main/java/org/checkerframework/checker/builder/TypesafeBuilderAnnotatedTypeFactory.java
```diff
@@ -554,6 +554,15 @@ private boolean hasAnnotation(Element element, Class<? extends Annotation> annot
     return element.getAnnotation(annotClass) != null;
   }
 
+  private boolean hasAnnotation(Element element, String annotName) {
+    return element.getAnnotationMirrors().stream()
```
msridhar: Can we use `AnnotationUtils.areSameByName` like in https://github.com/msridhar/returnsrecv-checker/pull/12?

kelloggm: Definitely. I guess I need more coffee this morning...

msridhar: 😆I only knew about that routine by chance

PR #46, opened 2019-07-25
Rename "typesafe builder checker" to "object construction checker"

mernst:
The package is still `builder` (that is, `org.checkerframework.checker.builder`).
Is it intentional that it was not renamed?

### DIFF
--- a/README-LOMBOK.md
+++ b/README-LOMBOK.md
@@ -16,8 +16,8 @@ Build the Object Construction Checker:
 ```
 git clone https://github.com/msridhar/returnsrecv-checker.git
 (cd returnsrecv-checker && ./gradlew build && ./gradlew publishToMavenLocal)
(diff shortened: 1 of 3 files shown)


kelloggm:
Yes, I intentionally did not rename the package, because I wasn't sure what else to call it - adding two levels of package names (`object.construction`) seems like it wouldn't fit with the naming convention; adding a dash between them (`object-construction`) makes it feel unwieldy; and neither seems to convey the right thing on its own (`object` seems to broad, and `construction` is vague).

Thanks for noticing and fixing the other missed instances!

mernst:
How about naming the package objectconstruction?  (You can't have a dash in a Java package name, though.)  I agree it's just a bit unwieldy, but it feels better than builder to me, and users will rarely write it.

kelloggm:
Sure, we can do that. It doesn't really matter, in the long run.

PR #54, opened 2019-07-26
Support for uploading artifacts to Maven Central

msridhar:
Analogous to msridhar/returnsrecv-checker#13


msridhar:
Note that this won't pass CI until msridhar/returnsrecv-checker#13 lands, but it passes tests locally

object-construction-qual/build.gradle
```diff
@@ -0,0 +1,41 @@
+plugins {
+    // Apply the java plugin to add support for Java
+    id 'java'
+
+    // For bringing in AWS dependencies. We're using the "legacy" version
+    // listed here: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-project-gradle.html,
+    // because the build fails when we use the Gradle 5+ way, saying that it can't find ec2.
+    // I believe that this is because we need the AWS SDK during testing, but not during the build.
+    id 'io.spring.dependency-management' version '1.0.7.RELEASE'
```
kelloggm: Why does the annotation package need the AWS SDK? That doesn't seem right - that should only be necessary for running the `DescribeImagesRequest` tests.

object-construction-qual/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+    id "com.diffplug.gradle.spotless" version "3.23.0"
+}
+
+sourceCompatibility = 1.8
+
+dependencies {
+    // This dependency is found on compile classpath of this component and consumers.
+    implementation 'org.checkerframework:checker:2.9.0'
```
kelloggm: This should also be `checker-qual`, because we don't want anyone who depends on the annotations to get the framework on their compile classpath.

I would also add a dependency on the annotation package for the returns receiver checker here, so that `@This` is on the classpath automatically without the need for a second annotation dependency (this is the current behavior).

object-construction-checker/build.gradle
```diff
@@ -57,40 +55,6 @@ tasks.withType(JavaCompile).all {
...
-            groupId = 'org.checkerframework'
-            artifactId = 'object-construction-qual'
-            version = "${tsbcVersion}"
-            artifact qualJar
-            pom.withXml {
-                def depNode = asNode().appendNode("dependencies").appendNode("dependency")
-                depNode.appendNode("groupId", "org.checkerframework")
-                depNode.appendNode("artifactId", "returnsrcvr-qual")
-                depNode.appendNode("version", "0.1-SNAPSHOT")
-            }
```
kelloggm: The code directly above is what currently makes the qual package depend directly on `returnsrcvr-qual`

msridhar:
Great comments.  I've updated the dependencies now so that anyone who depends on `object-construction-qual` will also get the returns receiver and Checker Framework qualifiers in their classpath.  This requires switching to the `java-library` plugin to use `api` dependencies.

msridhar:
I've confirmed that in the generated pom file for `object-construction-qual` the dependencies are now in the `compile` scope as desired:
```xml
  <dependencies>
    <dependency>
      <groupId>org.checkerframework</groupId>
      <artifactId>checker-qual</artifactId>
      <version>2.9.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>net.sridharan.objectconstruction</groupId>
      <artifactId>returnsrcvr-qual</artifactId>
      <version>0.1-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
```

PR #55, opened 2019-07-26
Only refine receivers that are not implicit member selects

kelloggm:
The fix for #44 was overzealous and introduced a new false positive bug. Code like the test added in this PR (`CompareTo.java`) causes a false positive like this (example is from a case study):

```
/Users/kelloggm/Research/lombok-testing/java-webauthn-server/webauthn-server-core/build/delombok/main/com/yubico/webauthn/data/PublicKeyCredentialDescriptor.java:94: error: [argument.type.incompatible] incompatible types in argument.
            return type.compareTo(other.type);
                                       ^
  found   : @CalledMethodsTop PublicKeyCredentialType
  required: @CalledMethods("compareTo") PublicKeyCredentialType
```

This PR fixes that oversight. I'm not sure that comparing on the `toString` is the best way to avoid the problem, but it seems to fix the problem. I thought about comparing on either the kind of tree or the type of receiver, but both are not exclusive to this problem, being `memberSelectTree`s and `FieldAccess`_s, respectively.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
@@ -71,7 +71,8 @@ public class ObjectConstructionTransfer extends CFTransfer {
     CFStore thenStore = result.getThenStore();
     CFStore elseStore = result.getElseStore();
 
-    while (receiver != null) {
+    while (receiver != null && !receiver.toString().startsWith("(this).")) {
+
       // Insert the new type computed previously as the type of the receiver.
       Receiver receiverReceiver = FlowExpressions.internalReprOf(atypefactory, receiver);
       thenStore.insertValue(receiverReceiver, newType);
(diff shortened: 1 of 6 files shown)


msridhar:
Hrm, I don't really understand what was broken before and how this fixes it.  Can you give a bit more explanation?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -71,7 +71,8 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
     CFStore thenStore = result.getThenStore();
     CFStore elseStore = result.getElseStore();
 
-    while (receiver != null) {
+    while (receiver != null && !receiver.toString().startsWith("(this).")) {
```
msridhar: Not sure how much we care or how much it matters in this case, but I think these `toString()` calls can be bad for checker performance

mernst: I agree we should avoid that.  String comparisons also tend to be brittle (such as depending on the number of parentheses).

mernst:
Please comment out the test, merge, and uncomment the test after Checker Framework version 3.0.0 is released with a fix for the underlying issue.

msridhar:
Is there a Checker Framework issue documenting the underlying problem?

mernst:
I don't see a Checker Framework issue documenting the underlying problem.

Here is the fix:
https://github.com/typetools/checker-framework/compare/master...smillst:enumTypeArgs?expand=1
The fix is delayed waiting for code review, so it won't be merged for about a week probably.

PR #56, opened 2019-07-27
Handle AutoValue Optionals

### LINKED ISSUE #52: Support Optional properties in AutoValue
If a property of an AutoValue object has any of the Optional types listed [here](https://github.com/google/auto/blob/4158a5fa71f1ef763e683b627f4d29bc04cfde9d/value/src/main/java/com/google/auto/value/processor/Optionalish.java#L38-L43) then the Builder property need not be set explicitly, as it will default to `Optional.absent()`.  We should not report errors for such properties.

msridhar:
Fixes #52

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -20,6 +20,7 @@ import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -608,4 +612,29 @@ private static String capitalize(String prop) {
    */
   private static final ImmutableSet<String> IGNORED_METHOD_NAMES =
       ImmutableSet.of("equals", "hashCode", "toString", "<init>", "toBuilder");
+
+  /** Taken from AutoValue source code */
+  private static final ImmutableSet<String> OPTIONAL_CLASS_NAMES =
+      ImmutableSet.of(
+          "com.".concat("google.common.base.Optional"), // subterfuge to foil shading
```
kelloggm: Do we really want to keep this in our code? I'm sure that they had a good reason to do it, but it is very confusing here.

msridhar:
Will land after #54 as CI job won't pass until that lands

PR #60, opened 2019-07-31
Support for NOT operation in @CalledMethodPredicate

kelloggm:
Supporting unsound bug-finding was surprisingly easy: the boolean expression parser we were using already supports the not operator, so the only code change I had to make was to changing the `isSubtype` implementation so top is not always assumed to be un-assignable: `@CalledMethodsUnknown` is a subtype of `@CalledMethodsPredicate("!foo")`, for example.

I also added two test cases. XOR isn't natively supported by our parser, but it's easily representable as `(a || b) && !(a && b)`, so that's not too bad.

I also added a note to the README that using the not operator in `CalledMethodsPredicate` makes the checker unsound.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,19 @@ automatically if it is on your compile classpath (as long as no annotation proce
 
 ## Specifying your code
 
-You can specify your code's contracts (what it expects from clients) by writing type annotations.
-A type annotation is written before a type.  For example, in `@NonEmpty List<@Regex String>`, `@NonEmpty` is a type annotation on `List`, and `@Regex` is a type annotation on `String`.
+The Object Construction Checker works as follows:
+ * It reads method specifications or contracts:  what they require when they are called.
+ * It keeps track of which methods have been called on each object.
+ * It warns if method arguments do not satisfy the method's specification.
+
+Most specifications are automatically inferred by the Object Construction
+Checker.  For example, it determines the specification of `build()` from
+`@Nullable` annotations, among other sources.
+
+You can specify your code's contracts (what it expects from clients) by
+writing type annotations.  A type annotation is written before a type.  For
+example, in `@NonEmpty List<@Regex String>`, `@NonEmpty` is a type
+annotation on `List`, and `@Regex` is a type annotation on `String`.
 
 The two most relevant annotations are:
 <dl>
--- a/README.md
+++ b/README.md
@@ -137,9 +137,9 @@ if (somePredicate) {
 }
 
 @CalledMethodsPredicate("! methodA") Object x;
-x = never;        // no warning
-x = oneBranch;    // no warning
-x = bothBranches; // warning
+x = never;        // no warning (methodA was never called)
+x = oneBranch;    // no warning (even though methodA might have been called)
+x = bothBranches; // warning (methodA was definitely called)
 ```
 
 Suppose that exactly one (but not both) of two methods should be called.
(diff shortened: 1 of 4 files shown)


object-construction-checker/tests/basic/Not.java
```diff
@@ -0,0 +1,61 @@
...
+    void callA(Foo f) {
+        f.a();
+    }
+
+    void test6(Foo f) {
+        callA(f);
+        // DEMONSTRATION OF UNSOUNDNESS
+        f.notA();
+    }
+
```
msridhar: Can we add a test case that demonstrates unsoundness due to a control-flow merge?  E.g., a case where `a()` is called the `then` branch but not the `else` branch

kelloggm: Done

mernst:
How hard would it be to support XOR rather than forcing users to use NOT?
I think XOR would be nicer to write.

README.md
```diff
@@ -68,6 +68,9 @@ objects such that:
 </dd>
 </dl>
 
+It is possible to use the `@CalledMethodsPredicate` for unsound bug-finding, rather than
+verification, by using the `!` (NOT) operator.
```
mernst: This text doesn't say what is unsound, nor how to use the checker for unsound bug-finding.
I think it would be clearer to say that "! x" means: "it isn't the case that x was definitely called" or "x was not called on every possible path" rather than "x is not called" or "x was not called on any path".
The checker is sound with respect to this semantics, right?
It might be helpful to say that there is not currently a way to express "x was not called".  That is more specific than the given text.

kelloggm:
@mernst the parser we are using does not support xor: https://github.com/spring-projects/spring-framework/issues/10922

If we wanted to support XOR, we could do one of these things:
* change parsers again. Our options are limited, because we want it to have a friendly license and support an explicit XOR operator. I'm not sure such a thing exists.
* roll our own boolean expression parser that does support XOR.
* use some kind of regular expression to replace an XOR token with the an equivalent expression that our existing parser can understand. I'm not sure that will be an easier than rolling our own parser, though.

I don't find any of these options appealing. Since `@CalledMethodsPredicate` annotations will mostly be written in stub files, not by end-users, I don't think the benefits outweigh the costs.

mernst:
Can you document the type hierarchy?  That is important to help users understand it.

kelloggm:
@mernst I have re-written the documentation for the `@CalledMethodsPredicate` annotation to make the type hierarchy clear.

msridhar:
> @mernst the parser we are using does not support xor: [spring-projects/spring-framework#10922](https://github.com/spring-projects/spring-framework/issues/10922)
> 
> If we wanted to support XOR, we could do one of these things:
> 
> * change parsers again. Our options are limited, because we want it to have a friendly license and support an explicit XOR operator. I'm not sure such a parser exists.
> * roll our own boolean expression parser that does support XOR.
> * use some kind of regular expression to replace an XOR token with the an equivalent expression that our existing parser can understand. I'm not sure that will be an easier than rolling our own parser, though.
> 
> I don't find any of these options appealing. Since `@CalledMethodsPredicate` annotations will mostly be written in stub files, not by end-users, I don't think the benefits outweigh the costs.

So I don't forget, another option here would be to contribute XOR support to SpEL.  I agree with the short-term decision not to support it.

PR #67, opened 2019-08-02
[WIP] Update to Checker Framework 2.10.0; fixes #66

### LINKED ISSUE #66: Upgrade to Checker Framework version 2.10.0
It would also be good to document how to do this.
(When I tried, I got a variety of obscure errors, but maybe I was doing something wrong.)

msridhar:
Fixes #66

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -1,6 +1,7 @@
 package org.checkerframework.checker.objectconstruction;
 
 import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableSet;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.NewClassTree;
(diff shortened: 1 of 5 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -1,6 +1,7 @@
 package org.checkerframework.checker.objectconstruction;
 
 import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableSet;
```
msridhar: @kelloggm @mernst this is weird.  Tests were failing until I added a Guava dependence and imported `ImmutableSet` from there.  Before, we were using `org.checkerframework.com.google.common.collect.ImmutableSet`.  But after updating Checker Framework from 2.9.0 to 2.10.0, invoking `ImmutableSet.of(...)` fails with this error:
```
java.lang.ClassNotFoundException: org.checkerframework.com.google.common.math.IntMath
```

Maybe something changed in how Checker Framework is packaged?  /cc @smillst

smillst: >  Before, we were using org.checkerframework.com.google.common.collect.ImmutableSet

Guava is only in checker.jar as a dependency of annotation-file-utilities-all.jar, which switched to the newest version https://github.com/typetools/annotation-tools/commit/27b51ae57177f8f3a3160dea8a5abbd5c9e96a1f.  Only part of Guava is used: https://github.com/typetools/annotation-tools/blob/27b51ae57177f8f3a3160dea8a5abbd5c9e96a1f/annotation-file-utilities/build.xml#L223-L229.

So, added a Guava dependence to your project is the right fix.

msridhar: 👍thanks @smillst!

PR #69, opened 2019-08-02
Further cleanup of Gradle build files

msridhar:
Consolidate some version numbers in the top-level build file to make upgrading easier

### DIFF
--- a/returnsrcvr-qual/build.gradle
+++ b/returnsrcvr-qual/build.gradle
@@ -14,7 +14,7 @@ sourceCompatibility = 1.8
 
 dependencies {
     // This dependency is found on compile classpath of this component and consumers.
-    implementation 'org.checkerframework:checker-qual:2.10.0'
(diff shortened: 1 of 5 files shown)


returnsrcvr-qual/build.gradle
```diff
@@ -14,7 +14,7 @@ sourceCompatibility = 1.8
 
 dependencies {
     // This dependency is found on compile classpath of this component and consumers.
-    implementation 'org.checkerframework:checker-qual:2.10.0'
```
mernst: The change implementation=>api doesn't seem related to the main point of this pull request, but OK.

msridhar: Yes true.  That was for consistency with `object-construction-qual` which makes this an api dependence.

msridhar:
I think for this kind of cleanup we can land without two reviews, so I'll go ahead

PR #80, opened 2019-08-07
Correct dispatching problem in toBuilder handling

kelloggm:
It seems that when I added support for Lombok `toBuilder` code I neglected to change the method called to produce the `@CalledMethods` annotation. This PR fixes that oversight.


kelloggm:
I spent some time trying to write such a test case, but I can't seem to find a minimal one.

This came up in a case study and I confirmed that this fix does fix the issue.

PR #81, opened 2019-08-08
Support for postconditions via @EnsuresCalledMethods

msridhar:
This PR adds support for postconditions by adding an `@EnsuresCalledMethods` annotation, leveraging Checker Framework's `@PostconditionAnnotation`.  This allows us to capture cases where an object is passed to some callee, and that callee invokes some of the required methods on the object.  See the `Postconditions` test for examples.


msridhar:
> Looks good to me.
> 
> It should be documented in the README with the other qualifiers, however. Can you add a few sentences there explaining it so that users can be aware?

Good point, just added some docs

Issue #58, opened 2019-07-30
NonNull fields with user-written defaults in Lombok builders


kelloggm:
Similar to #52, sometimes programmers write `@NonNull` on `Optional` types. Unlike AutoValue, Lombok has no built-in support for this, so there's no equivalent thing to do for Lombok.

However, we still get false positives like this:

```java
@lombok.Builder
public class Foo {
   @lombok.NonNull Optional<String> bar;

      public static class FooBuilder {
          private Optional<String> foo = Optional.empty();
     }
}
```

That is, the builder defaults the optional to being empty, and Lombok generates the rest - making the `@NonNull` annotation redundant.

This problem, however, is more general: a programmer could set a default in this manner even for non-optional fields, and we would issue a false positive expecting that field to be set.

PR #85, opened 2019-08-10
Add support to handle many defaulted properties in builders

kelloggm:
This is a partial fix for #58. It removes many of the false positives in my case study, but theoretically could miss some.

Here is the design:
* if it is possible to recover a tree from an element representing a field inside a builder, check if the tree has initializers. If so, treat the field as defaulted (the user must have written the initializer, because Lombok never generates one).
* keep a cache of elements that have been defaulted in this way, and consult it when the tree is unavailable. If there is an element in the cache, treat the element as defaulted. Elements are valid keys; other caches using them as keys exist, including e.g. [here](https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L381)

Obviously that cache might not have been filled, so there still might be false positives from #58. However, this change dramatically reduces them in practice, so I think closing the issue is in order.


msridhar:
Do we know why trees become unavailable in this case?  I thought that would happen due to annotation processor rounds, but with our delombok approach I didn't think that would happen here.

In any case these changes look reasonable to me.  I can approve the PR if needed.

kelloggm:
ASTs aren't available when the code comes from bytecode, I think.

PR #87, opened 2019-08-11
Documentation on Lombok/AutoValue defaulting

kelloggm:
Mike pointed out that we didn't have good docs for the Lombok/AutoValue defaulting rules.

Mike, are these docs sufficient to help you understand what is going on? If not, what other information do you need?

Manu, have I gotten the AutoValue rules correct? Are there any caveats like those I added for Lombok at the end that I don't know about?

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,13 +182,15 @@ However, if will not warn if there are some paths on which both methods are call
 
 The checker automatically inserts default annotations for code that uses builders generated
 by Lombok and AutoValue. There are three places annotations are usually inserted:
-* an `@CalledMethods` annotation is placed on the receiver of the `build()` method. For Lombok,
+* A `@CalledMethods` annotation is placed on the receiver of the `build()` method, capturing the
+setter methods that must be invoked on the builder before calling `build()`. For Lombok,
 this annotation's argument is the set of `@lombok.NonNull` fields that do not have default values.
-For AutoValue, it is the set of fields that are neither `Optional` nor `@Nullable`.
-* if the object has a `toBuilder()` method (for example, if the `toBuilder = true` option is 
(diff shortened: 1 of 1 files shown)


msridhar:
I am not sure it is super beneficial to document all the special cases we have handled for Lombok and AutoValue.  As an alternative, we could say that for the `build()` method, we add `@CalledMethods` for all the required setter calls, handling various special cases.  Isn't that basically the only thing that is "visible" to a user of the tool?  If we're writing documentation for developers building on the tool that is different.

We could also say that we may have missed some special cases and marked some `Builder` setters as required when in fact they are not, and that users should file an issue if they find such a case.

mernst:
I agree that brevity, and focus on what users care about, will lead to the most helpful documentation.

kelloggm:
I think it is important to have complete documentation of exactly what annotations we are placing in users' code automatically. This is especially important for Lombok, because experienced users often partially overwrite Lombok's generated code and insert their own versions. In order for such users to figure out how our tool will interact with their customizations, they'll need access to this information.

This text appears at the end of the README, and is linked earlier on for curious readers. It is otherwise there only for experienced users. Since our README is somewhat disorganized, that's only signaled by its placement at the end.

I am strongly of the opinion that this text needs to appear somewhere. I'm open to putting it into a "for advanced users" section or document, but none exists. Would that be preferable?

msridhar:
Sounds good @kelloggm. I don’t think devs do any hacking of AutoValue-generated code so the details are less important there. I’m fine with landing this, or I can make a pass on it later / tomorrow morning.

kelloggm:
This is just documentation, so it can wait for you to do a pass - there's no hurry.

Hacking on what Lombok generates is fairly common.

PR #89, opened 2019-08-13
Use Value Checker to specialize to withFilters(owner) case

kelloggm:
This change introduces specialized handling for a common false positive case that we've seen when running AMI sniping case studies. The `withFilters` method can be used to set an owner field, but the object construction checker, on its own, is unable to detect that an owner has been set, causing a false positive.

This change adds the Value Checker from the mainline Checker Framework as a subchecker, with a corresponding increase in execution time. Since this handling is so specialized, I'm not sure that this should actually be merged into `master`. However, it should help @martinschaef with his experiments. We should discuss whether merging this into `master` makes sense.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -11,13 +13,17 @@
...
 public class ObjectConstructionChecker extends BaseTypeChecker {
 
+  public static final String USE_VALUE_CHECKER = "useValueChecker";
+
   @Override
   protected LinkedHashSet<Class<? extends BaseTypeChecker>> getImmediateSubcheckerClasses() {
     LinkedHashSet<Class<? extends BaseTypeChecker>> checkers =
         super.getImmediateSubcheckerClasses();
     checkers.add(ReturnsRcvrChecker.class);
+    checkers.add(ValueChecker.class);
```
msridhar: Can we only add this sub-checker if the flag is set?

kelloggm: I don't think so. `BaseTypeChecker#getOptions` checks the options on all subcheckers, which requires calling this method (`getImmediateSubcheckerClasses`). So if I do that I get a stack overflow.

There might be a way to get around that restriction, though. I'll look into it.

msridhar: As a temporary hack we could use a JVM property rather than a Checker Framework option, if we need to change Checker Framework to allow checking of options here.

kelloggm: I've figured out a way to get the passed options list directly. I'm like 90% sure it's evil, though, so @mernst should probably look at it before we merge this.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -11,13 +13,19 @@
...
 public class ObjectConstructionChecker extends BaseTypeChecker {
 
+  public static final String USE_VALUE_CHECKER = "useValueChecker";
+
   @Override
   protected LinkedHashSet<Class<? extends BaseTypeChecker>> getImmediateSubcheckerClasses() {
     LinkedHashSet<Class<? extends BaseTypeChecker>> checkers =
         super.getImmediateSubcheckerClasses();
     checkers.add(ReturnsRcvrChecker.class);
+    if (this.processingEnv.getOptions().containsKey(USE_VALUE_CHECKER)) {
```
msridhar: This is nicer than the hack I was suggesting 😄

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -150,6 +156,47 @@ boolean returnsThis(final MethodInvocationTree tree) {
...
+          // find a constructor invocation, digging through receivers
+          while (filterTree != null && filterTree.getKind() == Tree.Kind.METHOD_INVOCATION) {
+            filterTree =
+                TreeUtils.getReceiverTree(((MethodInvocationTree) filterTree).getMethodSelect());
+          }
+          if (filterTree == null) {
+            continue;
+          }
+          if (filterTree.getKind() == Tree.Kind.NEW_CLASS) {
+            ExpressionTree argTree = ((NewClassTree) filterTree).getArguments().get(0);
```
mernst: I would find the name `withFiltersArg` (the argument to `withFilters`) clearer than `argTree`.
But, this and the following 10 lines feel like they should be abstracted out:  "the value is exactly 'owner'".  If there is not a method that does that, I suggest adding it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
mernst: This seems fine to me.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
kelloggm: Such a method would belong in the Value Checker, but I don't think it exists.

I could make a pull request to the Value Checker, but I think that would delay this PR unnecessarily. It's also not clear it would be useful elsewhere.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
mernst: Please add a brief comment about why you did it this way.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
mernst: Please make that independently, and add a brief comment here to use that after a release is made.

kelloggm: https://github.com/typetools/checker-framework/pull/2726

PR #90, opened 2019-08-14
Add test cases for generics support

msridhar:
Without the `@DefaultFor` meta-annotation on `@CalledMethodsTop` we get an inscrutable `return.type.incompatible` error for the new `Generics` test.  We ran into a similar issue with the returns receiver checker, and I stole the solution from there.

### DIFF
--- a/object-construction-checker/tests/basic/Generics.java
+++ b/object-construction-checker/tests/basic/Generics.java
@@ -3,7 +3,7 @@ import org.checkerframework.checker.returnsrcvr.qual.*;
 
 class Generics {
 
-    private <T extends Symbol> T getMember(Class<T> type) {
+    private <@CalledMethods("isStatic") T extends Symbol> T getMember(Class<T> type) {
         T sym = getMember(type);
         if (sym != null && sym.isStatic()) {
             return sym;
--- a/object-construction-checker/tests/basic/Generics.java
+++ b/object-construction-checker/tests/basic/Generics.java
@@ -1,0 +1,19 @@
+import org.checkerframework.checker.objectconstruction.qual.*;
+import org.checkerframework.checker.returnsrcvr.qual.*;
+
+class Generics {
+
+    private <T extends Symbol> T getMember(Class<T> type) {
+        T sym = getMember(type);
+        if (sym != null && sym.isStatic()) {
+            return sym;
+        }
+        return null;
+    }
+
+    static interface Symbol {
+
+        boolean isStatic();
+
+    }
+}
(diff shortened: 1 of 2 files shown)


msridhar:
Here is what @smillst wrote regarding a similar problem for returns receiver:

> The standard defaulting rules set the qualifier of type variable lower bounds to bottom. (https://checkerframework.org/manual/#climb-to-top)  Returns receiver checker should change this default to top, then that error will go away. (Add the meta-annotation @DefaultFor(LOWER_BOUND) to the top annotation class.) This will make types like `List<@This Object>` illegal, but that seems ok to me.

We do use generic support for one benchmark, so let me check that things still work there before landing this change.

msridhar:
Actually this breaks generics support for that other benchmark; need to think more about this change.

msridhar:
Do we have any idea how to get this code to type check with our checker?
```java
class Generics {
     private <T extends Symbol> T getMember(Class<T> type, boolean b) {
        if (b) {
            T sym = getMember(type, !b);
            if (sym != null && sym.isStatic()) {
                return sym;
            }
        } else {
            T sym = getMember(type, b);
              if (sym != null) {
                return sym;
              }
        }
        return null;
    }
     static interface Symbol {
         boolean isStatic();
     }
}
```

As it stands I get an error like (with verbose generics):
```
tests/basic/Generics.java:10: error: [return.type.incompatible] incompatible types in return.
                return sym;
                       ^
  type of expression: T[ extends @CalledMethods("isStatic") Symbol super @CalledMethods("isStatic") Void]
  method return type: T[ extends @CalledMethodsTop Symbol super @CalledMethodsBottom Void]
:-1: other: tests/basic/Generics.java:15: error: [return.type.incompatible] incompatible types in return.
                return sym;
                       ^
  type of expression: T[ extends @CalledMethodsTop Symbol super @CalledMethodsTop Void]
  method return type: T[ extends @CalledMethodsTop Symbol super @CalledMethodsBottom Void]
```
I don't understand these error messages; in both cases it looks like the `type of expression` is within the bounds of the return type.

@kelloggm @smillst @mernst

smillst:
> I don't understand these error messages; in both cases it looks like the type of expression is within the bounds of the return type.

It's not enough for it to be with in the bounds.  For one use of a type variable to be a subtype of another use, its bounds must match exactly or it's upper bound must be below the lower bound of the supertype.

You can fix this warning:
```
private <@CalledMethodsTop T extends Symbol> T getMember(Class<T> type, boolean b) {
```

or
```
private <T extends Symbol> @CalledMethodsTop T getMember(Class<T> type, boolean b) {
```

msridhar:
Thanks, @smillst!  This PR now just adds some tests of our generics support, with no changes to the checker.

PR #93, opened 2019-08-14
Custom error messages for Lombok/AutoValue-style finalizers

kelloggm:
This PR replaces the ugly and difficult to understand `method.invocation.invalid` errors of the form:

```
tests/lombok/BuilderTest.java:17: error: [method.invocation.invalid] call to build() not allowed on the given receiver.
        BuilderTest.builder().x(0).build(); // bad builder
                                        ^
  found   : @CalledMethods("x") BuilderTestBuilder
  required: @CalledMethods("y") BuilderTestBuilder
```

with nicer errors of the form:

```
tests/lombok/BuilderTest.java:17: error: [finalizer.invocation.invalid] This finalizer cannot be invoked, because the following methods have not been called: y() 
        BuilderTest.builder().x(0).build(); // bad builder
                                        ^
```

The primary benefit of these new error messages is that they name the methods that actually need to be called explicitly, and don't list all the methods that already have been called - so for complex builders the user no longer needs to diff the two lists in their head.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -20,4 +26,75 @@
     checkers.add(ReturnsRcvrChecker.class);
     return checkers;
   }
+
+  /** Parses the String representation of an @CalledMethods annotation into */
```
msridhar: Incomplete Javadoc?

object-construction-checker/tests/lombok/LombokBuilderExample.java
```diff
@@ -0,0 +1,82 @@
+// Generated by delombok at Wed Aug 14 15:57:15 PDT 2019
```
msridhar: What does this test have to do with the overall PR?

kelloggm: Sorry, I forgot to document that. I wanted a testcase with two different required methods that weren't being called so that I could make sure I got the parsing right, and didn't have one. I figured I ought to commit it, even if it was mostly duplicative; we can remove it if we want. Other than the extra required argument it's nearly identical to tests we already have.

kelloggm:
@msridhar can you cut a release with these changes whenever you get a chance? I think @martinschaef would like to be able to use them in the user study

PR #95, opened 2019-08-16
Handle build method not named build() for AutoValue

msridhar:
The build method for an AutoValue Builder need not be named `build`: https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#build_names

So we now match the build method based on type, not name.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -278,7 +278,7 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
 
       if (builderKind != BuilderKind.NONE) {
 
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -292,6 +292,19 @@ assert hasAnnotation(nextEnclosingElement, AutoValue.class)
...
+    private boolean isBuilderBuildMethod(
+        ExecutableElement element, BuilderKind builderKind, Element nextEnclosingElement) {
+      switch (builderKind) {
+        case LOMBOK:
+          return "build".equals(element.getSimpleName().toString());
+        case AUTO_VALUE:
+          // return type should be enclosing AutoValue class
+          return TypesUtils.getTypeElement(element.getReturnType()).equals(nextEnclosingElement);
+        default:
+          throw new RuntimeException("not possible!");
```
kelloggm: Can you make this exception more descriptive?

PR #96, opened 2019-08-16
Properly handle inherited methods for AutoValue and AutoValue Builder types

msridhar:
See the new `Inheritance` test case.  Before, we would not consider inherited abstract methods when determining required properties or potential setters in a Builder class.  To fix, we now look at abstract methods in all supertypes of `@AutoValue` and `@AutoValue.Builder` classes.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -313,10 +313,10 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
   }
 
   /**
-   * Is e a setter for a builder?
+   * Is e a setter for an AutoValue builder?
    *
    * @param member member of builder or one of its supertypes
-   * @param builderElement element for the builder
+   * @param builderElement element for the AutoValue builder
    * @return {@code true} if e is a setter for the builder, {@code false} otherwise
    */
   private boolean isAutoValueBuilderSetter(Element member, Element builderElement) {
@@ -330,6 +330,7 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
     }
     TypeMirror retType = ((ExecutableElement) member).getReturnType();
     if (retType.getKind().equals(TypeKind.TYPEVAR)) {
+      // instantiate the type variable for the Builder class
       retType =
           AnnotatedTypes.asMemberOf(
                   getContext().getTypeUtils(),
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -313,15 +317,31 @@ private boolean isBuilderBuildMethod(
    *     Immutable type
    */
   private Set<String> getAllBuilderSetterMethodNames(Element builderElement) {
-    return builderElement.getEnclosedElements().stream()
+    AnnotatedTypeMirror builderType = getAnnotatedType(builderElement);
+    return getAllSupertypes((Symbol) builderElement).stream()
+        .flatMap(e -> e.getEnclosedElements().stream())
         .filter(
             e -> {
```
kelloggm: This lambda is long enough that I think it ought to be its own method so that you can name and document it. As-is, this method is really hard to read.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -455,40 +475,52 @@ private void handleToBuilderType(
...
-            requiredPropertyNames.add(name);
-          }
-        }
-      }
-    }
-    return requiredPropertyNames;
+    return getAllSupertypes((Symbol) autoValueClassElement).stream()
+        .flatMap(e -> e.getEnclosedElements().stream())
+        .filter(
+            member -> {
```
kelloggm: This lambda is even more difficult to read - it's nearly a full page of code. I also think this should be its own method, with documentation describing the property it is computing.

msridhar:
Good feedback; lambdas have been extracted.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -308,22 +312,48 @@ private boolean isBuilderBuildMethod(
     }
   }
 
+  /**
+   * Is e a setter for an AutoValue builder?
```
kelloggm: I think `e` is supposed to be `member` here and in the rest of this comment.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -308,22 +312,48 @@ private boolean isBuilderBuildMethod(
...
+    // either the return type should be the builder itself, or it should be a Guava immutable type
+    return isGuavaImmutableType(retType)
+        || builderElement.equals(TypesUtils.getTypeElement(retType));
+  }
   /**
    * @param builderElement Element for a Lombok or AutoValue builder
    * @return names of all methods whose return type is the builder itself or returns a Guava
    *     Immutable type
    */
   private Set<String> getAllBuilderSetterMethodNames(Element builderElement) {
```
kelloggm: Can you also change the name and documentation of this method to indicate that it only works for AutoValue now?

PR #97, opened 2019-08-16
Ensure AutoValue build method is abstract

msridhar:
Sometimes AutoValue Builders have a non-abstract `build` method that does some validation, delegating to some other abstract method (usually named `autoBuild`) for the generated build logic.  We should only add a `@CalledMethods` annotation on the abstract method.

### DIFF
--- a/object-construction-checker/tests/autovalue/Validation.java
+++ b/object-construction-checker/tests/autovalue/Validation.java
@@ -1,0 +1,44 @@
+import com.google.auto.value.AutoValue;
+import org.checkerframework.checker.objectconstruction.qual.*;
+import org.checkerframework.checker.nullness.qual.*;
+
+
+@AutoValue
+abstract class Validation {
+
+  public abstract String name();
+
+  static Builder builder() {
+    return new AutoValue_Validation.Builder();
+  }
+
+  @AutoValue.Builder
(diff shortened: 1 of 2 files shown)


object-construction-checker/tests/autovalue/Validation.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @AutoValue.Builder
+  abstract static class Builder {
+
+    abstract Builder setName(String name);
+
+    abstract Validation autoBuild();
+
+    public Validation build(@CalledMethods("setName") Builder this) {
```
kelloggm: I don't understand why this `@CalledMethods` annotation shouldn't be inferred.

msridhar: Now we infer a `@CalledMethods` annotation for `autoBuild()`, since that's the method whose implementation gets generated by AutoValue.  We can't infer anything for this other public `build()` method since the developer wrote the code and we don't know what it does without analyzing the code.  In principle, it could set all the properties itself, in which case `@CalledMethodsTop` is a fine type for its receiver.  Make sense?

kelloggm: Yes, thanks for explaining.

PR #100, opened 2019-08-17
Fix error message parsing bug

### LINKED ISSUE #99: Incorrect methods listed in error message from checker
Test case:
```java
@AutoValue
abstract class SetInsideBuild {
  public abstract String name();
  public abstract int size();
  static Builder builder() {
    return new AutoValue_SetInsideBuild.Builder();
  }
  @AutoValue.Builder
  abstract static class Builder {
    abstract Builder setName(String name);
    abstract Builder setSize(int value);
    abstract SetInsideBuild autoBuild();
    public SetInsideBuild build(Builder this) {
      return this.setSize(4).autoBuild(); // error here
    }
  }
}
```
I would expect an error on the marked line indicating that `setName()` may not have been called.  Instead, the tool reports:
```
error: [finalizer.invocation.invalid] This finalizer cannot be invoked, because the following methods have not been called: setName() setSize() 
      return this.setSize(4).autoBuild();
```
This is pretty confusing 😄 I confirmed that I get no error message if I add an annotation to fix `build()`:
```java
    public SetInsideBuild build(@CalledMethods("setName") Builder this) {
      return this.setSize(4).autoBuild();
    }
```
So I'm pretty sure this is an issue in printing the error message.  I'm not sure how to write a test for the exact message using the Checker Framework testing tools.

kelloggm:
I was stripping quotes from around required methods, but not actual methods, so the String comparison didn't work.

Fixes #99.

The new test case is the one from #99; it doesn't fail before this, since exact error messages aren't checked, but I figured there was no harm in including it. I tested that the new error messages are right by manually inspecting the output.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
@@ -49,24 +49,28 @@ public class ObjectConstructionChecker extends BaseTypeChecker {
       return Collections.emptySet();
     }
 
-    String withoutCMAnno = calledMethodsString.trim().substring("@CalledMethods({".length());
+    String withoutCMAnno = calledMethodsString.trim().substring("@CalledMethods(".length());
     String withoutBaseType = withoutCMAnno.substring(0, withoutCMAnno.lastIndexOf(' '));
-    if (withoutBaseType.length() >= 2) {
-      String[] withQuotes = withoutBaseType.substring(0, withoutBaseType.length() - 2).split(", ");
-      Set<String> result = new HashSet<>();
-      for (String s : withQuotes) {
-        if (s.startsWith("\"")) {
-          s = s.substring(1);
-        }
-        if (s.endsWith("\"")) {
-          s = s.substring(0, s.length() - 1);
-        }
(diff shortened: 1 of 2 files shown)


kelloggm:
@msridhar after you review this, can you cut a release and let me know? Thanks!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -53,8 +52,18 @@
     String withoutCMAnno = calledMethodsString.trim().substring("@CalledMethods({".length());
```
msridhar: @kelloggm one confusing thing is that when you have one called method, the annotation looks like `@CalledMethods("setSize")`, but when you have multiple, it looks like `@CalledMethods({"setSize",...})`.  Since we always trim `"@CalledMethods({"` from the beginning I worry that this is working by accident.  Could you tweak the logic or at least add some comments explaining how both cases are handled?

kelloggm: I've rewritten the method to hopefully make it much more clear - it wasn't quite working by accident, but it was a mess. Thanks for pushing me to clean it up.

PR #101, opened 2019-08-19
Improve precision of Value Checker enabled AMI sniping mode

kelloggm:
This PR improves the precision of the `useValueChecker` option when used for AMI sniping detection. In particular, it makes the following improvements, each of which is based on a real false positive we have encountered:
* add `setExecutableUsers` and `withExecutableUsers` to the whitelist in the stub file
* add three new permitted filters: `image-id`, `owner-alias`, `owner-id`
* treat calls to `withName` as the same as using the filter constructor
* introspect into calls to `Collections.singletonList` and examine their arguments

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -201,8 +201,17 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
             if (valueType.hasAnnotation(StringVal.class)) {
               AnnotationMirror valueAnno = valueType.getAnnotation(StringVal.class);
               List<String> possibleValues = getValueOfAnnotationWithStringArgument(valueAnno);
-              if (possibleValues.size() == 1 && "owner".equals(possibleValues.get(0))) {
-                return "withOwners";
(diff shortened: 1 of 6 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -163,47 +169,80 @@ boolean returnsThis(final MethodInvocationTree tree) {
...
+        // passing through as many method invocation trees as needed. This code is searching
+        // for code of the form:
+        // new Filter("owner").withValues("...")
+        // or code of the form:
+        // new Filter().*.withName("owner").*
+
+        // Set to non-null iff a call to withName was observed; in that case, this variable's
+        // value is the argument to withName.
+        String withNameArg = null;
+        ValueAnnotatedTypeFactory valueATF = getTypeFactoryOfSubchecker(ValueChecker.class);
```
msridhar: I wonder if this would be cleaner if written as a visitor, with some top-level checks on the type of the expression before applying the visitor?  Not urgent, possibly for a future refactoring

kelloggm: Maybe. I think the complexity here is still manageable, but if we were to make it much more complex I would agree with you.

For now, though, I'm going to leave it as-is.

PR #104, opened 2019-09-03
Update to Checker Framework 2.11.0

msridhar:
This allows the checker to run atop Error Prone javac.  Besides the version bump, some Gradle changes were required to place the Error Prone javac in the bootclasspath when running tests.


msridhar:
Update: let's hold off on landing this until I test the `README` instructions for running the checker and make sure we understand the implications for users.

msridhar:
Another status update.  If we update to Checker Framework 2.11.0, the easiest way for our users to consume the checker will be via the Checker Framework Gradle plugin, so the Error Prone javac gets auto-injected.  But, that plugin enables individual checks via the `javac` `-processor` flag, which disables processor auto-discovery (see https://github.com/kelloggm/checkerframework-gradle-plugin/issues/20).  I think this will be a real pain for users on AutoValue to deal with, as AutoValue is bundled with something like five annotation processors.  I feel like we should hold off on this update until we have a plan for simpler support for AutoValue users (or projects like https://github.com/google/nomulus that use both AutoValue and Dagger via auto-discovery).

msridhar:
@kelloggm this should be ready for review now.  Note that I have removed the separate `README-LOMBOK.md` file as it became mostly redundant with the main README.  I integrated the remaining Lombok-specific instructions into the main `README.md` file.

README.md
```diff
@@ -33,25 +33,42 @@ annotationProcessor() ...`.
...
 
-1. Make your Maven/Gradle project depend on `net.sridharan.objectconstruction:object-construction-checker:0.1.5`.
-
-  For example, for Gradle, add the following to the `build.gradle` file (adding the entries to the extant `repositories` and `dependencies` blocks if present):
+1. Add the [org.checkerframework](https://github.com/kelloggm/checkerframework-gradle-plugin) Gradle plugin to the `plugins` block of your `build.gradle` file:
 
+  ```groovy
+  plugins {
+      ...
+      id "org.checkerframework" version "0.4.0"
```
kelloggm: Maybe we should add a note here about using the latest version of the gradle plugin? It is released frequently so it would be good to avoid having `0.4.0` directly in this README.

kelloggm: We could change this to `0.4.+` instead of `0.4.0`, so that it will use the latest version in the `0.4.x` series. I don't actually know if that works for plugin versions, but it does for regular gradle versions so I assume so.

msridhar: Unfortunately that doesn't work for plugin versions; dynamic versions are not supported there.  Let me know if you want me to add a note encouraging the user to check for a more recent plugin version.

PR #108, opened 2019-10-14
Separate framework checker for returns receiver checker

ranmanli:
The framework support code has been separated and moved to another package named framework. 
Each framework implements interface `FrameworkSupport` and method `knownToReturnThis`, which checks if the method is in the framework and returns this. Returns receiver checker goes through every framework support and adds `@This` annotation if `knownToReturnThis` returns true.


returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/AutoValueSupport.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.checkerframework.checker.returnsrcvr.ReturnsRcvrAnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.javacutil.TypesUtils;
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * 
+ * AutoValue support for returns receiver checker, check if the current class or the super class is AutoValue Builder
```
msridhar: Grammar: "AutoValue support for returns receiver checker. Checks if the current class or its super class is an AutoValue Builder"

returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/FrameworkSupport.java
```diff
@@ -0,0 +1,13 @@
+package org.checkerframework.checker.framework;
+
+
+import org.checkerframework.checker.returnsrcvr.ReturnsRcvrAnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+
+public interface FrameworkSupport {
```
msridhar: Add JavaDoc for this interface

returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/FrameworkSupport.java
```diff
@@ -0,0 +1,13 @@
...
+	/**
+	 * 
+	 * @return {@code true} if the method is generated by framework and returns {@code this} 
+	 */
+	boolean knownToReturnThis(AnnotatedTypeMirror.AnnotatedExecutableType t, ReturnsRcvrAnnotatedTypeFactory typeFactory);
```
msridhar: Let's not take the `ReturnsRcvrAnnotatedTypeFactory` as a parameter here.  Instead, either make the relevant methods in `ReturnsRcvrAnnotatedTypeFactory` into static methods, or move them into their own class for utility methods.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
```diff
@@ -54,63 +63,40 @@ public Void visitExecutable(AnnotatedTypeMirror.AnnotatedExecutableType t, Void
...
-    }
+  }
+  
+  private boolean isConstructor(AnnotatedTypeMirror.AnnotatedExecutableType t) {
+	  ExecutableElement element = t.getElement();
+	  return element.getKind().equals(ElementKind.CONSTRUCTOR);
   }
 
-  private boolean hasAnnotation(Element element, Class<? extends Annotation> annotClass) {
+  public boolean hasAnnotation(Element element, Class<? extends Annotation> annotClass) {
```
msridhar: These two methods should be `static`.  Or, @kelloggm, do you know if there are built-in methods in Checker Framework we can use instead of these?

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
```diff
@@ -19,15 +22,21 @@
...
 
   AnnotationMirror THIS_ANNOT;
+  Collection<FrameworkSupport> frameworkSupports;
 
   public ReturnsRcvrAnnotatedTypeFactory(BaseTypeChecker checker) {
     super(checker);
     THIS_ANNOT = AnnotationBuilder.fromClass(elements, This.class);
+    
+    //add supports for frameworks
+    frameworkSupports = new ArrayList<FrameworkSupport>();
```
msridhar: @kelloggm instead of fixing this list here, we'd like to take in a list as a parameter to the constructor, or something equivalent.  Then, with a single command-line flag in the object-construction checker we could control the framework support there and in this checker.  However, it looks like the object construction checker currently instantiates this sub-checker via reflection.  I think we should go ahead and land this change once it's ready but we may need your help to figure out how to thread through the configuration properly

msridhar:
@ranmanli to fix the Travis failure run `./gradlew spotlessApply` locally to fix any formatting issues

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
kelloggm: I don't know of such a convenience method in the framework itself. I agree these should be static.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/FrameworkSupport.java
kelloggm: I suggest moving those methods into a non-instantiable utility class, in the style of the `AnnotationUtils` class in the Checker Framework proper.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/LombokSupport.java
```diff
@@ -0,0 +1,35 @@
...
+import org.checkerframework.javacutil.TypesUtils;
+
+/**
+ * 
+ * Lombok support for returns receiver checker
+ *
+ */
+public class LombokSupport implements FrameworkSupport {
+	
+    public boolean knownToReturnThis(AnnotatedTypeMirror.AnnotatedExecutableType t, ReturnsRcvrAnnotatedTypeFactory typeFactory) {
```
kelloggm: Here and elsewhere, methods that override a method from a super type should be annotated with `@Override`

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
kelloggm: I agree we should land this change as soon as it is ready, and deal with threading that through when we're ready.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
ranmanli: I tried to make them static but it seems they are using the non-static variables `elements` from `AnnotatedTypeFactory`, so I can't make them static.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/FrameworkSupport.java
ranmanli: It's actually caused by the following problem, I didn't find the way to make `hasAnnotation` static.

ranmanli: @kelloggm I am not sure about what you suggest, like have two methods `knownToReturnThisInLombok` and `knownToReturnThisInAutoValue` in a `FrameworkSupportUtils` non-instantiable class?

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
```diff
@@ -55,61 +62,37 @@ public Void visitExecutable(AnnotatedTypeMirror.AnnotatedExecutableType t, Void
...
+  }
+
+  public boolean hasAnnotation(Element element, Class<? extends Annotation> annotClass) {
     return elements.getAllAnnotationMirrors(element).stream()
         .anyMatch(anm -> AnnotationUtils.areSameByClass(anm, annotClass));
   }
 
-  private boolean hasAnnotationByName(Element element, String annotClassName) {
+  public boolean hasAnnotationByName(Element element, String annotClassName) {
     return elements.getAllAnnotationMirrors(element).stream()
```
msridhar: So we only use `elements` to call `getAllAnnotationMirrors()` which handles inherited annotations.  But for the use cases here I don't think we care about inherited annotations.  So, we should be able to replace the call as below and then make the enclosing method `static`:
```suggestion
     return element.getAnnotationMirrors().stream()
```

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
msridhar: See below for suggestions on how to make the methods static

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
```diff
@@ -55,61 +62,37 @@ public Void visitExecutable(AnnotatedTypeMirror.AnnotatedExecutableType t, Void
...
   }
 
-  private boolean hasAnnotation(Element element, Class<? extends Annotation> annotClass) {
+  private boolean isConstructor(AnnotatedTypeMirror.AnnotatedExecutableType t) {
+    ExecutableElement element = t.getElement();
+    return element.getKind().equals(ElementKind.CONSTRUCTOR);
+  }
+
+  public boolean hasAnnotation(Element element, Class<? extends Annotation> annotClass) {
     return elements.getAllAnnotationMirrors(element).stream()
```
msridhar: To enable `static`:
```suggestion
    return element.getAnnotationMirrors().stream()
```

ranmanli: I made them static in this way, and move them both to a utility class named "FrameworkSupportUtils". Also don't need to pass AnnotatedTypeFactory as parameter anymore.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
```diff
@@ -19,15 +18,21 @@
 import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
 import org.checkerframework.javacutil.AnnotationBuilder;
 import org.checkerframework.javacutil.AnnotationUtils;
-import org.checkerframework.javacutil.TypesUtils;
 
 public class ReturnsRcvrAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
   AnnotationMirror THIS_ANNOT;
+  Collection<FrameworkSupport> frameworkSupports;
```
msridhar: Can we JavaDoc this field?

msridhar:
@kelloggm ping on this

PR #109, opened 2019-11-02
Update to Checker Framework 3.0.0

msridhar:
Had to make the following changes:

* Add Error Prone javac as a `compileOnly` build dependence for the checkers, since the Checker Framework jar no longer includes those classes.  I chose `compileOnly` since in any valid runtime configuration those classes should be present.
* Deleted a couple of all-systems tests that are no longer valid.

Note: this is still only tested on JDK 8.  We should support JDK 11 next, at which point further configuration changes may be required.


msridhar:
@kelloggm once we get JDK 11 working as well we should probably update https://github.com/typetools/templatefora-checker as well

Issue #110, opened 2019-11-15
Handle AutoValue Extensions


msridhar:
AutoValue has extensions that can affect which methods of an `@AutoValue` type correspond to properties.  See, e.g, the following comment:

https://github.com/google/auto/blob/4543619f73dc94cc9a7232f340067e5bfe8ddb24/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java#L193-L198

We need to find a way to make our AutoValue support robust to extensions "consuming" abstract methods like `Parcelable.describeContents()`.  Right now, we get a crash because we see an abstract method like `describeContents()` but no corresponding setter in the builder.

/cc @lazaroclapp

msridhar:
More details on why this is hard.  AutoValue Extensions can "consume" arbitrary abstract methods of the `@AutoValue` class, so that AutoValue doesn't generate a concrete implementation of them:

https://github.com/google/auto/blob/389355d065fbe189d8f7a388beeaebeaaf1d13ae/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java#L383

To know which methods are getting consumed in the current build, we need to know what each extension is doing.  But, the extensions are run by AutoValue, which may run before or after our checker.  My feeling is the best hope we have is to run the extensions ourselves to see what is getting consumed, but that would take some significant investigation / implementation, and it might not even work.

PR #111, opened 2019-11-16
Don't fail when we can't find an AutoValue Builder setter method

msridhar:
Before, we would fail with a `RuntimeException` if we couldn't match what we thought was an AutoValue property to a corresponding setter in the builder.  But, we don't currently handle AutoValue extension (see #110), and adding full and proper handling will be pretty non-trivial.  So, instead of failing for these cases, just tolerate them.

Add a test case using the AutoValue Parcel extension to expose the problem, along with some stub Android classes so we don't need to pull in a full Android dependence.


msridhar:
Sorry for the confusion. The test passes because this PR disables the check for a matching builder setter. The test fails on current master with a RuntimeException. I will update comments to be clearer

PR #114, opened 2019-12-23
Add flag to disable framework supports

ranmanli:
Added a command-line flag to disable the framework supports for both object construction checker and returnsrcvr checker. To use it, specify the framework supports we want to disable in a comma-separated list to the flag `disabledFrameworkSupports`. For example, to disable both AutoValue and Lombok support, use `-AdisabledFrameworkSupports=AutoValue,Lombok`. I also added a test `disabledframeworkTest` with two simple test cases adapted from _UseAutoValueBuilder.java_ and _LombokBuilderExample.java_ but removed all `// :: error:` comments. The code passed the test by specifying the flag as `-AdisabledFrameworkSupports=AutoValue,Lombok`.


README.md
```diff
@@ -211,6 +211,10 @@ passed to Lombok's `@Builder` annotation), then the return type of that method i
 the same `@CalledMethods` annotation as the receiver of `build()`, using the same rules as above.
 * A `@This` annotation is placed on the return type of each setter in the builder's implementation.
 
+You can disable the framework supports by specifying them in a comma-separated list to the 
+command-line flag `disabledFrameworkSupports`.  For example, to disable both Lombok and AutoValue supports,
+use `diabledFrameworkSupports=AutoValue,Lombok` .
```
kelloggm: There is a typo in the name of the command line option.

Also, this is an option to an annotation processor, so just supplying `-disabledFrameworkSupports=AutoValue`, for example, will not work; it has to be `-AdisabledFrameworkSupports=AutoValue`.

object-construction-checker/tests/disabledframeworks/README.md
```diff
@@ -0,0 +1,2 @@
+These two test cases are adapted from _autovalue/UseAutoValueBuilder.java_ and _lombok/LombokBuilderExample.java_
+but removed all `// :: error:` comments.
```
kelloggm: Can you add a sentence here with the reasoning? That is, this README should say that these are testcases for disabling framework support explicitly.

README.md
```diff
@@ -211,6 +211,10 @@ passed to Lombok's `@Builder` annotation), then the return type of that method i
 the same `@CalledMethods` annotation as the receiver of `build()`, using the same rules as above.
 * A `@This` annotation is placed on the return type of each setter in the builder's implementation.
 
+You can disable the framework supports by specifying them in a comma-separated list to the 
+command-line flag `disabledFrameworkSupports`.  For example, to disable both Lombok and AutoValue supports,
+use `-AdiabledFrameworkSupports=AutoValue,Lombok` .
```
kelloggm: Both of the arguments need to be corrected. They should be `disableFrameworkSupports` and `-AdisableFrameworkSupports=AutoValue,Lombok`, respectively.

returnsrcvr-checker/src/main/java/org/checkerframework/checker/returnsrcvr/ReturnsRcvrAnnotatedTypeFactory.java
```diff
@@ -29,11 +29,29 @@ public ReturnsRcvrAnnotatedTypeFactory(BaseTypeChecker checker) {
     super(checker);
     THIS_ANNOT = AnnotationBuilder.fromClass(elements, This.class);
 
-    // add supports for frameworks
-    frameworkSupports = new ArrayList<FrameworkSupport>();
-    frameworkSupports.add(new AutoValueSupport());
-    frameworkSupports.add(new LombokSupport());
+    boolean disableAutoValue = false;
```
kelloggm: All of this code is shared between the Object Construction Checker and Returns Receiver Checker. Can we factor it out into a common static method in `FrameworkSupport`? That method would take as input the `String` that's currently in the local variable `disabledFrameworkSupports`, and return a list of `FrameworkSupport`s. Then this code duplication isn't necessary.

ranmanli: Thanks for the review. The framework supports for object construction checker and returnsrcvr checker are different. Maybe I can factor them out separately and put the methods into the corresponding `FrameworkSupportUtils` class?

kelloggm: Good point. I think that your plan makes sense.

That said, the core logic to decide which framework supports to use (i.e. the parsing/toUpperCase code) should be shared.

One design would be to make an enum with the values AUTO_VALUE and LOMBOK, and then return that in the common code. Then, the code in each checker could convert the shared enum into an appropriate list of `FrameworkSupport`s.

ranmanli: For the common logic part, I am wondering where should I put it into? I can try to factor it out into a static method in class `ReturnsRcvrChecker`, which is also the place I put the flag name and framework names.

msridhar: @ranmanli I think `FrameworkSupportUtils` inside returns receiver checker is a good place for this code.  You could follow the design @kelloggm suggests and have a nested enum there, and then a static method that takes the `String` from the option and returns an [`EnumSet`](https://docs.oracle.com/javase/7/docs/api/java/util/EnumSet.html) of which frameworks are specified.  Then each checker could go through the set and disable support for the corresponding frameworks.

README.md
```diff
@@ -211,6 +211,10 @@ passed to Lombok's `@Builder` annotation), then the return type of that method i
 the same `@CalledMethods` annotation as the receiver of `build()`, using the same rules as above.
 * A `@This` annotation is placed on the return type of each setter in the builder's implementation.
 
+You can disable the framework supports by specifying them in a comma-separated list to the 
+command-line flag `disableFrameworkSupports`.  For example, to disable both Lombok and AutoValue supports,
+use `-AdiableFrameworkSupports=AutoValue,Lombok` .
```
kelloggm: There is *still* a typo here - a missing `s` in `diable`

returnsrcvr-checker/src/main/java/org/checkerframework/checker/framework/FrameworkSupportUtils.java
```diff
@@ -10,6 +14,31 @@
   // this class is non-instantiable
   private FrameworkSupportUtils() {}
 
+  public enum Framework {
+    AUTO_VALUE,
+    LOMBOK;
+  }
+
+  public static Set<Framework> getFrameworkSet(String option) {
```
kelloggm: This method is missing javadoc

PR #120, opened 2020-01-29
Add subtyping between CalledMethodsPredicate annotations

kelloggm:
Based on a reviewer's suggestion in the paper. This is a rare case, but should resolve that TODO nicely and makes the design much more elegant.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -424,13 +424,16 @@ public boolean isSubtype(final AnnotationMirror subAnno, final AnnotationMirror
 
       if (AnnotationUtils.areSameByClass(subAnno, CalledMethodsPredicate.class)) {
         if (AnnotationUtils.areSameByClass(superAnno, CalledMethodsPredicate.class)) {
```
mernst: This does not seem to handle the case of comparing a `@CalledMethods` annotation with a `@CalledMethodsPredicate` annotation.  Please see the tests I added.

kelloggm: You're right, it didn't. I added that case by turning the `@CalledMethods` annotation into a conjunct and passing it to the solver.

kelloggm:
@mernst this is ready for your review again

msridhar:
My only slight concern here is taking the dependence on the SMT solver.  Does it have a lot of transitive deps?  Having lots of dependencies can occasionally lead to weird issues with annotation processors that users have a hard time working around, e.g., if another processor depends on a different version of some library.  So I generally try to minimize 3rd-party dependencies for annotation processors.

kelloggm:
There are two possibly-dangerous dependencies in the dependency tree for the solver:
* Guava, which it has set to 28.0-jre. That's probably fine - everything already depends on Guava anyway.
* `org.checkerframework:checker-qual:2.5.2`, which is a very old version of the qualifier libraries. That's presumably only a dependency because they have annotations, so using a later version should be fine.

Javaparser's only dependency is on the JDK.

msridhar:
Ok, sounds good.  One other question.  Since we are pulling in an SMT solver, can we get rid of this dependency?
```groovy
    // For the @CalledMethodsPredicate evaluation
    implementation "org.springframework:spring-expression:5.1.7.RELEASE"
```
There might be downsides to using the SMT solver for this (ugly, slower, etc.).  Just thought I'd ask.

kelloggm:
Probably. I'll look into it and make a second PR if it makes sense.

README.md
```diff
@@ -146,10 +146,15 @@ subtyping.  Then
...
-of the Object Construction Checker.)  For this reason, programmers usually only
-write `@CalledMethodsPredicate` annotations on formal parameters.
+Subtyping between two `@CalledMethodsPredicate` annotations is determined by checking
+whether the proposed subtype implies the proposed supertype. In
+particular:
+
+`CalledMethodsPredicate(`P`) T1` &#8849; `CalledMethodsPredicate(`Q`) T2` iff T1 &#8849; T2
+and &#0172; (P &#8658; Q) is unsatisfiable.
+
+If either P or Q contain operators other than `&&` or `||`, the result of the comparison
```
mernst: It would be better behavior for the checker to issue a warning to the user, rather than to silently behave in an unexpected way.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -15,6 +29,118 @@ public CalledMethodsPredicateEvaluator(final Set<String> cmMethods) {
     this.cmMethods = cmMethods;
   }
 
+  /**
+   * Return true iff the boolean formula "lhs -> rhs" is valid, treating all variables as
+   * uninterpreted.
+   *
+   * <p>Used to determine subtyping between two @CalledMethodsPredicate annotations that are not
```
mernst: It also handles `@CalledMethodsPredicate` versus `@CalledMethods`.  I suggest just dropping this paragraph, since it's obvious from the code and can easily get out of date.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -15,6 +29,118 @@ public CalledMethodsPredicateEvaluator(final Set<String> cmMethods) {
...
+   * identical.
+   */
+  public static boolean implies(final String lhsOrig, final String rhsOrig) {
+
+    JavaParser parser = new JavaParser();
+
+    String preamble = "class DUMMY { boolean DUMMY() { return ";
```
mernst: It's a bit weird that the flow of code in this method is:
 * parsing-related
 * set up solver
 * more parsing-related

Please move all the parsing-related code together.  Once doing that, it becomes clear that lines 43-54 should be merged into `compilationUnitToBooleanFormula` to avoid duplication.  (That method might need to be slightly renamed.)

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -15,6 +29,118 @@ public CalledMethodsPredicateEvaluator(final Set<String> cmMethods) {
...
+    String afterward = "; } }";
+
+    String lhs = preamble + lhsOrig + afterward;
+    String rhs = preamble + rhsOrig + afterward;
+
+    CompilationUnit lhsAst = parser.parse(lhs).getResult().orElse(null);
+    CompilationUnit rhsAst = parser.parse(rhs).getResult().orElse(null);
+
+    if (lhsAst == null || rhsAst == null) {
+      return false;
```
mernst: Returning `false` is surprising behavior and does not lead itself to a good debugging experience.  I think the checker should issue a warning so that the user knows there is a syntax error.  Otherwise, it could be very difficult to determine that there is a parenthesis missing, or `&` should have been `&&`, etc.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -15,6 +29,118 @@ public CalledMethodsPredicateEvaluator(final Set<String> cmMethods) {
...
+
+    BlockStmt theBlock = ast.getType(0).getMembers().get(0).asMethodDeclaration().getBody().get();
+
+    com.github.javaparser.ast.expr.Expression theExpression =
+        theBlock.getStatements().get(0).asReturnStmt().getExpression().get();
+
+    try {
+      return expressionToBooleanFormula(theExpression, booleanFormulaManager);
+    } catch (UnsupportedOperationException e) {
+      throw e;
```
mernst: What is the point of this vacuous `catch` clause?  I suggest removing it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -15,6 +29,118 @@ public CalledMethodsPredicateEvaluator(final Set<String> cmMethods) {
...
+    }
+  }
+
+  /** Actual implementation of conversion between Javaparser expression and boolean formula. */
```
mernst: "Actual implementation of" is not very informative.  Maybe the comment would be better off without it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -15,6 +29,118 @@ public CalledMethodsPredicateEvaluator(final Set<String> cmMethods) {
...
+            expressionToBooleanFormula(
+                theExpression.asBinaryExpr().getLeft(), booleanFormulaManager),
+            expressionToBooleanFormula(
+                theExpression.asBinaryExpr().getRight(), booleanFormulaManager));
+      }
+    } else if (theExpression.isEnclosedExpr()) {
+      return expressionToBooleanFormula(
+          theExpression.asEnclosedExpr().getInner(), booleanFormulaManager);
+    }
+    throw new UnsupportedOperationException();
```
mernst: This should be more informative; at the least, it should pass back either `theExpression` or a string representation of it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: The checker already issues an error for unparseable formulae in `ObjectConstructionVisitor`.

kelloggm: (and there's no other way to create a predicate formula besides writing it by hand, or a conjunction of pure literals which is known to be parseable)

kelloggm:
@mernst I've addressed your comments. I think I could refactor this even more to make the interface cleaner between the rest of the checker and the evaluator, but I'm going to wait on that until I do a PR that removes the SPEL dependency, which will be a follow-up to this one.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
mernst: Thanks for that information.  Can you note that in a comment?  In that case, I think it would be better to you change the `return false` to throwing an exception, to indicate that it cannot happen.

kelloggm: I changed it to throw an informative exception

README.md
```diff
@@ -146,10 +146,15 @@ subtyping.  Then
...
-of the Object Construction Checker.)  For this reason, programmers usually only
-write `@CalledMethodsPredicate` annotations on formal parameters.
+Subtyping between two `@CalledMethodsPredicate` annotations is determined by checking
+whether the proposed subtype implies the proposed supertype. In
+particular:
+
+`CalledMethodsPredicate(`P`) T1` &#8849; `CalledMethodsPredicate(`Q`) T2` iff T1 &#8849; T2
+and &#0172; (P &#8658; Q) is unsatisfiable.
+
+If either P or Q contains operators other than `&&` or `||`, the result of the comparison
```
mernst: Is this still true?  If the input is invalid, I think it's better to issue an error than to continue.  I suggest changing the behavior (or removing the statement if it is not true).

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -450,14 +450,23 @@ public boolean isSubtype(final AnnotationMirror subAnno, final AnnotationMirror
...
+            return CalledMethodsPredicateEvaluator.implies(subPredicate, superPredicate);
+          }
+        } else if (AnnotationUtils.areSameByClass(superAnno, CalledMethods.class)) {
+          // If the supertype is a called methods type, treat it as a conjunction of its
+          // arguments and use the predicate evaluator.
+          List<String> superVals = getValueOfAnnotationWithStringArgument(superAnno);
+          String superValPredicate = String.join(" && ", superVals);
+          return CalledMethodsPredicateEvaluator.implies(subPredicate, superValPredicate);
         }
         return false;
```
mernst: I find it difficult to follow the logic of this method -- to verify that all the cases are handled.  Why is there a `return false;` here as a fallthrough to the possibilities of `superAnno` being `CalledMethodsPredicate` and `CalledMethods`, but there is none below at line 487 where the two cases are assumed to be exhaustive?  I would expect the two cases to be parallel.

Above, why do lines 442-446 have an else that returns false, but line 449 does not?  Again, I would expect the two cases to be parallel.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -1,23 +1,181 @@
...
+  /** Construct a solver to use when comparing predicates. Each query should use a new solver. */
+  private static SolverContext setupSolver() {
+    Configuration config = Configuration.defaultConfiguration();
+    LogManager log = LogManager.createNullLogManager();
+    ShutdownNotifier notifier = ShutdownManager.create().getNotifier();
+    try {
+      return SolverContextFactory.createSolverContext(
+          config, log, notifier, SolverContextFactory.Solvers.SMTINTERPOL);
+    } catch (InvalidConfigurationException e) {
+      return null;
```
mernst: This is guaranteed to lead to a null pointer exception later, which will be harder to debug.
More generally, the code suppresses a number of exceptions and I think it would be better to at least log them, to help users understand unexpected behavior.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -1,23 +1,181 @@
...
+      String formula, BooleanFormulaManager booleanFormulaManager) {
+
+    String classWithFormula = "class DUMMY { boolean DUMMY() { return " + formula + "; } }";
+
+    CompilationUnit ast = parser.parse(classWithFormula).getResult().orElse(null);
+
+    if (ast == null) {
+      throw new BugInCF(
+          "Encountered an unparseable formula while parsing an @CalledMethodsPredicate "
+              + "annotation, but ObjectConstructionVisitor failed to stop compilation. Unparseable formula: "
```
mernst: The fact about ObjectConstructionVisitor is an implementation detail.  You could write a code comment here, but I would keep the error message short (just the last, brief sentence).

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -1,23 +1,181 @@
...
+            expressionToBooleanFormula(
+                theExpression.asBinaryExpr().getLeft(), booleanFormulaManager),
+            expressionToBooleanFormula(
+                theExpression.asBinaryExpr().getRight(), booleanFormulaManager));
+      }
+    } else if (theExpression.isEnclosedExpr()) {
+      return expressionToBooleanFormula(
+          theExpression.asEnclosedExpr().getInner(), booleanFormulaManager);
+    } else if (theExpression.isUnaryExpr()) {
+      if (theExpression.asUnaryExpr().getOperator().equals(UnaryExpr.Operator.LOGICAL_COMPLEMENT)) {
```
mernst: The documentation states that only && and || are supported, not !.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -1,23 +1,181 @@
...
+   * @param trueLiterals the literals in the boolean expression to treat as "true"
+   * @return whether the expression evaluates to true in the context where only the literals in
+   *     trueLiterals are true, and all other literals are false
+   */
+  public static boolean evaluate(String expression, Collection<String> trueLiterals) {
 
-    for (String cmMethod : cmMethods) {
+    for (String cmMethod : trueLiterals) {
       expression = expression.replaceAll(cmMethod, "true");
     }
```
mernst: On line 182, `_` is also a legal character in a Java identifier.  However, "0-9" are not legal as the first character.
On line 182, you need to use word delimiters "\b" around the symbol being replaced.
On line 186, you also need word delimiters, but the hack might not be necessary if you use the correct regular expression in the replacement.

Please add tests for `evaluate`.  I don't see any in the repository.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: The documentation says that `!` is permitted both in the `README` and on the public entry-point that calls this method (`implies`). What documentation are you referring to?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: I added a few, but this isn't the point of this PR. I also fixed the regex to match the actual Java identifier rules, with the exception of unicode support, which SPEL (the expression parser we're using) doesn't have. I added a commented-out test for that.

kelloggm: Also, I do not need `\b` delimiters, and they're actively harmful: we want to replace e.g. `a` in `!a` or `(a || b)`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
mernst: Sorry, my mistake.

What I should have said was that the documentation says that all symbols are treated as uninterpreted, but it looks like `false` and `true` might be specially treated.  I'm not sure about that, though.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
mernst: I agree that these fixes could be in a separate PR, since they are unrelated.  I wanted to point them out, though.

What is wrong with using `\b` delimiters to avoid matching in the middle of an identifier (which is the cause of the "tfalse" bug)?  Can you give a specific example?

kelloggm: If I add `\b` delimiters, the test case `Not.java` starts to fail with three unexpected `method.invocation.invalid` errors. A representative failure is line 14:

```java
class Not {

    class Foo {
        void a() {}
        void notA(@CalledMethodsPredicate("!a") Foo this) {}
    }

    void test1(Foo f) {
        f.notA(); // ERROR HERE
    }
} 
```

kelloggm: SPeL throws an exception (partial call stack):

```
:-1: other: error: SourceChecker.typeProcess: unexpected Throwable (SpelEvaluationException) while processing tests/basic/CmPredicate.java; message: EL1007E: Property or field 'a' cannot be found on null
  Compilation unit: tests/basic/CmPredicate.java
  Last visited tree at line 3 column 1:
  class CmPredicate {
  Exception: org.springframework.expression.spel.SpelEvaluationException: EL1007E: Property or field 'a' cannot be found on null; Stack trace: org.springframework.expression.spel.ast.PropertyOrFieldReference.readProperty(PropertyOrFieldReference.java:213)
  org.springframework.expression.spel.ast.PropertyOrFieldReference.getValueInternal(PropertyOrFieldReference.java:104)
  org.springframework.expression.spel.ast.PropertyOrFieldReference.getValueInternal(PropertyOrFieldReference.java:91)
  org.springframework.expression.spel.ast.SpelNodeImpl.getValue(SpelNodeImpl.java:185)
  org.springframework.expression.spel.ast.OpOr.getBooleanValue(OpOr.java:56)
  org.springframework.expression.spel.ast.OpOr.getValueInternal(OpOr.java:47)
  org.springframework.expression.spel.ast.OpOr.getValueInternal(OpOr.java:37)
  org.springframework.expression.spel.ast.SpelNodeImpl.getTypedValue(SpelNodeImpl.java:114)
  org.springframework.expression.spel.standard.SpelExpression.getValue(SpelExpression.java:173)
  org.checkerframework.checker.objectconstruction.CalledMethodsPredicateEvaluator.evaluate(CalledMethodsPredicateEvaluator.java:194)
```

mernst: What change did you make?  It's impossible to diagnose without seeing the change you made.  You could check it into a branch.

kelloggm: There are several of these, in multiple test cases (both the `BasicTest` and the `EC2Test`). I printed out some of the error-triggering code, and every instance had at least one character other than the identifier in the predicate, like `!a` or `(a`.

mernst: Try running this:
```
public class ReplaceTest {
  public static void main(String[] args) {
    String expression = "(a && abcde && true && false)";
    System.out.println(expression);
    System.out.println(expression.replaceAll("((?!true)[_a-zA-Z][_a-zA-Z0-9]*)", "false"));
    System.out.println(expression.replaceAll("((?!true)\\b[_a-zA-Z][_a-zA-Z0-9]*\\b)", "false"));
  }
}
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: Only `evaluate` treats `true` and `false` as boolean literals. `implies` and its implementation, of which this is part, really do treat all names as uninterpreted. I suspect it will actually fail if a `true` literal is passed, because Javaparser won't turn that into a `NameExpression`, so we'll reach the fall-through case.

mernst: OK, that sounds good.  I was assuming they treated boolean literals in the same way.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: (I also filed an issue wrt `evaluate` and unicode: https://github.com/kelloggm/object-construction-checker/issues/126)

kelloggm: > Try running this:
> 
> ```
> public class ReplaceTest {
>   public static void main(String[] args) {
>     String expression = "(a && abcde && true && false)";
>     System.out.println(expression);
>     System.out.println(expression.replaceAll("((?!true)[_a-zA-Z][_a-zA-Z0-9]*)", "false"));
>     System.out.println(expression.replaceAll("((?!true)\\b[_a-zA-Z][_a-zA-Z0-9]*\\b)", "false"));
>   }
> }
> ```

Be that as it may, making the same change in `CalledMethodsPredicateEvaluator` causes the tests to fail. Because it's unrelated to the substance of this PR, please file an issue and we can address it later.

kelloggm: > What change did you make? It's impossible to diagnose without seeing the change you made. You could check it into a branch.

I checked it in on the `subtyping-for-cmp-with-new-regex` branch

mernst: The problem is that you didn't use the regex that I used above, but enclosed it in "(...)*" which let it match empty strings.  I fixed the `subtyping-for-cmp-with-new-regex` branch to use the right regex and it passes CI.

kelloggm: Thanks for fixing this - I knew there must have been some kind of misunderstanding.

msridhar:
I'd like to cut a new release before this change lands.  I plan to do so tonight.  I will comment again when that is done

msridhar:
Ok, I've cut release 0.1.7 and pushed it to Maven Central

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -1,31 +1,176 @@
...
+  }
+
+  /** Converts a Java boolean expression into a boolean formula in the SMT solver's format. */
+  private static BooleanFormula formulaStringToBooleanFormula(
+      String formula, BooleanFormulaManager booleanFormulaManager) {
+
+    String classWithFormula = "class DUMMY { boolean DUMMY() { return " + formula + "; } }";
 
-    expression = expression.replaceAll("((?!true)[a-zA-Z0-9])+", "false");
+    CompilationUnit ast = parser.parse(classWithFormula).getResult().orElse(null);
```
mernst, line 102: Rather than the next 10 lines, use `StaticJavaParser.parseExpression`.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
```diff
@@ -1,31 +1,176 @@
...
+   *
+   * @param expression a boolean formula in Java format, as a String
+   * @param trueVariables the variables in the boolean expression to treat as "true"
+   * @return whether the expression evaluates to true in the context where only the variables in
+   *     trueVariables are true, and all other variables are false
+   */
+  public static boolean evaluate(String expression, Collection<String> trueVariables) {
+
+    for (String cmMethod : trueVariables) {
+      expression = expression.replaceAll("\\b" + cmMethod + "\\b", "true");
```
mernst, line 170: Rather than these string replacements before parsing, would it be better to do the substitution after parsing?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: I had apparently overlooked that method. Thanks - this code is now much cleaner!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/CalledMethodsPredicateEvaluator.java
kelloggm: I don't understand the suggestion. The way that this method works is to convert all names in the input expression into either `true` or `false` literals (depending on whether they are in `trueVariables` or not), and then evaluate the resulting expression.

It's a somewhat moot point regardless, though, because I will need to rewrite this code anyway to avoid the dependency on `SpelExpressionParser` before we can move this code into the framework. I'll re-evaluate the question when I'm doing that work.

PR #122, opened 2020-02-04
Alias old annotations to fully support Lombok's `checkerframework = true` option

kelloggm:
I copied the delombok'd test code directly from Lombok's source, so this should match exactly what they do.

The downside is that we have to go back to releasing annotations in the `org.checkerframework.checker.builder.qual` package, but I don't think that's a deal breaker. I considered marking them as deprecated, but that caused extra warnings when processing delombok'd code so I decided against it.

The annotations this adds support for are:
* `@CalledMethods` in the `builder.qual` namespace, which behaves identically to `@CalledMethods`
* `@NotCalledMethods`, which they (speculatively) generate, does nothing
* `@ReturnsReceiver` is trusted if it's found but no `@This` annotation is. This is the only one I'm a little worried about, but I think it's probably fine. The Javadoc on it makes it clear that users should use `@This` instead, and that's also what our README etc say

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -19,6 +19,8 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
+
+import org.checkerframework.checker.builder.qual.ReturnsReceiver;
 import org.checkerframework.checker.framework.FrameworkSupportUtils;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.checker.objectconstruction.framework.AutoValueSupport;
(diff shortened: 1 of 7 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -150,7 +167,20 @@ boolean returnsThis(final MethodInvocationTree tree) {
     AnnotatedTypeMirror methodATm = rrATF.getAnnotatedType(methodEle);
     AnnotatedTypeMirror rrType =
         ((AnnotatedTypeMirror.AnnotatedExecutableType) methodATm).getReturnType();
-    return rrType != null && rrType.hasAnnotation(This.class);
+    if (rrType != null && rrType.hasAnnotation(This.class)) {
```
msridhar: Just write this `if` condition as an or?

returnsrcvr-qual/src/main/java/org/checkerframework/checker/builder/qual/ReturnsReceiver.java
```diff
@@ -0,0 +1,24 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This declaration annotation indicates that the method on which it is written returns exactly the
+ * receiver object.
+ *
+ * <p>This annotation can only be written on a method declaration. It is inherited by all overrides
+ * of that method.
```
msridhar: Does the new test case also test the override functionality?  Not sure if it matters

kelloggm: It does not. I also don't think it really matters, but I can add a test.

kelloggm:
Wrt to the lombok thing - I tried to reproduce it with their actual release and couldn't. As far as I can tell, the issue is in their internal "verbose" delombok mode, which isn't accessible from their actual release. I can still report it, but it's kind of silly without a reproducible example since the only non-compilable files are their own tests.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
kelloggm: Good thought. Done.

msridhar:
Sorry I meant to get Lombok to generate `@This` and our other “modern”
annotations

kelloggm:
Yeah, I can make a PR to that effect. I'll wait to submit it until we have a new OCC release, though.

PR #139, opened 2020-03-07
Improved handling of AutoValue autoBuild() methods

### LINKED ISSUE #129: Better support for autoBuild() in AutoValue
AutoValue has a convention for doing additional validation of property values, via use of an `autoBuild()` method in addition to the usual `build()` method.  Here's the documentation:

https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#validate

Right now, our checker automatically injects a `@CalledMethods` annotation for the `autoBuild()` method in such cases, but *not* for the `build()` method.  The thinking behind this was that the developer can write arbitrary code in the `build()` method and we cannot know what annotation is the correct one.  The problem here is that if developers are following the convention from the docs, we get false positive warnings, since the `@CalledMethods` annotation for required properties is missing from `build()`.  We end up with a false positive warning at the `autoBuild()` call.  This comes up several times in experiments for our ICSE paper.

I now think we should use a different strategy.  We should auto-generate a `@CalledMethods` annotation for `build()` that is the same as the one we add for `autoBuild()`.  This is the correct thing to do if developers are following the convention.  If they do not follow the convention (e.g., by initializing properties within `build()` before calling `autoBuild()`), we should allow the developer to write their own `@CalledMethods` annotation on `build()`, which we do not override with our auto-generated annotation.  I think this will lead to the "right thing" happening in nearly all cases.

Nargeshdb:
it injects a @CalledMethods annotation for both autoBuild() and build() methods.

Fixes #129

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -57,7 +57,7 @@ import org.checkerframework.javacutil.TreeUtils;
 public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
   /** The top annotation. Package private to permit access from the Transfer class. */
-  final AnnotationMirror TOP;
+  public final AnnotationMirror TOP;
 
   /** The bottom annotation. Package private to permit access from the Transfer class. */
   final AnnotationMirror BOTTOM;
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -57,7 +57,7 @@ import org.checkerframework.javacutil.TreeUtils;
 public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
   /** The top annotation. Package private to permit access from the Transfer class. */
-  public final AnnotationMirror TOP;
+  final AnnotationMirror TOP;
 
   /** The bottom annotation. Package private to permit access from the Transfer class. */
   final AnnotationMirror BOTTOM;
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
@@ -105,7 +105,8 @@ public class AutoValueSupport implements FrameworkSupport {
           getAutoValueRequiredProperties(nextEnclosingElement, avBuilderSetterNames);
       AnnotationMirror newCalledMethodsAnno =
           createCalledMethodsForAutoValueProperties(requiredProperties, avBuilderSetterNames);
-      // only add the new @CalledMethods annotation if there is not already a @CalledMethods annotation present
+      // only add the new @CalledMethods annotation if there is not already a @CalledMethods
+      // annotation present
       AnnotationMirror possibleBuildAnnotations =
           t.getReceiverType()
               .getAnnotationInHierarchy(
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
@@ -20,15 +20,12 @@ import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import org.checkerframework.checker.objectconstruction.ObjectConstructionAnnotatedTypeFactory;
-import org.checkerframework.checker.objectconstruction.qual.CalledMethods;
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
 import org.checkerframework.framework.util.AnnotatedTypes;
-import org.checkerframework.framework.util.AnnotationMirrorMap;
 import org.checkerframework.javacutil.AnnotationUtils;
 import org.checkerframework.javacutil.TreeUtils;
 import org.checkerframework.javacutil.TypesUtils;
-import org.checkerframework.org.apache.bcel.classfile.ElementValue;
 
 /**
  * AutoValue Support for the Object Construction Checker, which adds CalledMethods annotation to the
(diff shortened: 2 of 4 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -57,7 +57,7 @@
 public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
   /** The top annotation. Package private to permit access from the Transfer class. */
-  final AnnotationMirror TOP;
+  public final AnnotationMirror TOP;
```
msridhar: @kelloggm we now need to access this top type from a different package.  We could change this public field to an accessor method.  But is there a cleaner way in general to access the top type of a type system?

mernst: If you can access the [QualifierHierarchy](https://checkerframework.org/api/org/checkerframework/framework/type/QualifierHierarchy.html), it has methods to get the top annotation.

kelloggm: For an annotation without arguments like `@CalledMethodsTop`, though, I think exposing it as a public final field on the AnnotatedTypeFactory is fine. Some of the internal Checker Framework checkers do that, for instance: https://github.com/typetools/checker-framework/blob/master/checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java#L100

mernst: I agree that is a more direct way.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -7,12 +7,7 @@
 import com.sun.tools.javac.code.Types;
 import com.sun.tools.javac.processing.JavacProcessingEnvironment;
 import java.beans.Introspector;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
msridhar: Can we get rid of this wildcard import?  You can disable wildcard imports in IntelliJ by following these instructions:

https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

Then if you do Code -> Optimize Imports in IntelliJ it will re-expand the import I think.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -76,8 +71,9 @@ public boolean isBuilderBuildMethod(ExecutableElement element) {
           : "class " + nextEnclosingElement.getSimpleName() + " is missing @AutoValue annotation";
       // it is a build method if it is an abstract method that returns the type with the @AutoValue
       // annotation
-      if (element.getModifiers().contains(Modifier.ABSTRACT)
-          && TypesUtils.getTypeElement(element.getReturnType()).equals(nextEnclosingElement)) {
+      if (
+      /*element.getModifiers().contains(Modifier.ABSTRACT)
```
msridhar: Please delete the commented-out code

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -106,7 +102,11 @@ public void handlePossibleBuilderBuildMethod(AnnotatedExecutableType t) {
           getAutoValueRequiredProperties(nextEnclosingElement, avBuilderSetterNames);
       AnnotationMirror newCalledMethodsAnno =
           createCalledMethodsForAutoValueProperties(requiredProperties, avBuilderSetterNames);
-      t.getReceiverType().addAnnotation(newCalledMethodsAnno);
+      AnnotationMirror possibleBuildAnnotations =
```
msridhar: Add a code comment here indicated that we only add the new `@CalledMethods` annotation if there is not already a `@CalledMethods` annotation present

object-construction-checker/tests/autovalue/SetInsideBuild.java
```diff
@@ -14,9 +14,24 @@ static Builder builder() {
...
-            return this.setSize(4).autoBuild(); // error here
+        public SetInsideBuild build(@CalledMethods({"setName"}) Builder this) {
+
+            return this.setSize(4).autoBuild();
         }
     }
+
+    public static void buildSomethingWrong() {
+        Builder b = builder();
+        //b.setGet("Frank");
```
msridhar: Delete this commented line

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -76,8 +76,7 @@ public boolean isBuilderBuildMethod(ExecutableElement element) {
           : "class " + nextEnclosingElement.getSimpleName() + " is missing @AutoValue annotation";
       // it is a build method if it is an abstract method that returns the type with the @AutoValue
```
msridhar: Please update the comment and remove "is an abstract method that"

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -106,7 +105,13 @@ public void handlePossibleBuilderBuildMethod(AnnotatedExecutableType t) {
           getAutoValueRequiredProperties(nextEnclosingElement, avBuilderSetterNames);
       AnnotationMirror newCalledMethodsAnno =
           createCalledMethodsForAutoValueProperties(requiredProperties, avBuilderSetterNames);
-      t.getReceiverType().addAnnotation(newCalledMethodsAnno);
+      // only add the new @CalledMethods annotation if there is not already a @CalledMethods
```
msridhar: 👍

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -106,7 +105,13 @@ public void handlePossibleBuilderBuildMethod(AnnotatedExecutableType t) {
...
+      // annotation present
+      AnnotationMirror possibleBuildAnnotations =
+          t.getReceiverType()
+              .getAnnotationInHierarchy(
+                  atypeFactory.getQualifierHierarchy().getTopAnnotation(newCalledMethodsAnno));
+      if (possibleBuildAnnotations == null) t.getReceiverType().addAnnotation(newCalledMethodsAnno);
```
kelloggm: It's considered bad style in Java code to use a one-line `if` statement, without braces. Please rewrite this to
```
      if (possibleBuildAnnotations == null) {
          t.getReceiverType().addAnnotation(newCalledMethodsAnno);
      }
```

Nargeshdb:
> to

Nargeshdb:
> Looks good to me, other than one minor formatting nit.

I fixed the formatting nit. Thanks.

msridhar:
@Nargeshdb looks like you closed by accident. Reopening

PR #156, opened 2020-04-14
Add note about lombok requirements

kelloggm:
Most users won't care, but anyone wanting to integrate this into a build system needs to know.

I also augmented the cf gradle plugin version to the latest, since I noticed it hadn't been updated in awhile.


kelloggm:
Good point about the CF version. That's just documentation - the tests don't actually use, so I'll revert that part of the change for now.

PR #164, opened 2020-05-11
Use new CF extension point to avoid needing to parse CM annotations for error reporting

### LINKED ISSUE #140: Error message formatting does not work with Checker Framework master branch
Our logic for adding nicer error message formatting with the `finalizer.invocation.invalid` error type no longer works when using the master branch of the Checker Framework.  We should keep this in mind when updating to the next Checker Framework release.

Unfortunately I don't think we have a regression test checking for the error message formatting.  @kelloggm is there any way to add such a test?

/cc @mernst

kelloggm:
Fixes #140

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
@@ -61,6 +61,10 @@ public class ObjectConstructionVisitor
     return super.visitMethodInvocation(node, p);
   }
 
+  /**
+   * Adds special reporting for method.invocation.invalid errors to turn them into
+   * finalizer.invocation.invalid errors.
+   */
   @Override
   protected void reportMethodInvocabilityError(
       MethodInvocationTree node, AnnotatedTypeMirror found, AnnotatedTypeMirror expected) {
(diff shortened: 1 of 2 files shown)


msridhar:
Do we match on the exact error message in any of our tests, to make sure this works?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -54,4 +60,31 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
     }
     return super.visitMethodInvocation(node, p);
   }
+
+  @Override
```
msridhar, line 64: Can we add Javadoc for this method, like we had before for the `reportError()` override?

kelloggm: Good point; I put it back.

kelloggm:
> Do we match on the exact error message in any of our tests, to make sure this works?

We do not, but I tested it by hand and the output looks the same. We do test that the error code is `finalizer.invocation.invalid`, though, so the only thing that could go wrong is the error message itself.

mernst:
> It would be great if we could test the actual error message somehow. But that may not be possible with standard CF testing infrastructure at the moment

To do so, you would write a jtreg test.  There are tests in the CF repository like that.

msridhar:
> To do so, you would write a jtreg test. There are tests in the CF repository like that.

Ok.  As far as I can see, this functionality is not exposed in the `framework-test` artifact.  But when we move the checker over to the Checker Framework repository we should add such a test.

PR #183, opened 2020-07-08
Always call checker

Nargeshdb:
This PR is our first version of AlwaysCall checker. It addressed three different cases where an object might lose its reference: 
1.  If a local variable that references an object goes out of scope, then the AlwaysCall condition is checked for the local variable right before going out of scope (Note that if a variable doesn’t exist in the successor store or if the next block is an exit block and the variable is not returned, it means the variable is going out of scope).
2.  If there is an assignment to a local variable that references to an object, then it checks AlwaysCall condition for the local variable node right before the assignment node.
3.  If an expression that has @AlwaysCall on its declared type isn’t stored in a local variable. 
The first two cases are handled inside the ObjectConstructionAnnotatedTypeFactory and the third case is handled inside the ObjectConstructionVisitor. 

For now, ownership is never transferred to formal parameters or object fields.  This could lead to false positives.

Cases are not handled yet:
-  It does not yet handle a sequence of call methods like:
makeFoo().a() or makeFoo().b().a()
where class Foo has @AlwaysCall(“a”) annotation, makeFoo() returns new Foo(), and b() returns @This.
-  It does not yet traverse exceptional successors, so it is unsound in the presence of exceptions.


msridhar:
@Nargeshdb can you please write a description of what exactly this PR does?

Nargeshdb:
> @Nargeshdb can you please write a description of what exactly this PR does, what cases are handled / not handled, etc.?

The description is updated.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
     return null;
   }
 
+  @Override
+  public void postAnalyze(ControlFlowGraph cfg) {
+    checkCFG(cfg.getEntryBlock(), cfg);
+    super.postAnalyze(cfg);
+  }
+
+  private void checkCFG(SpecialBlock entry, ControlFlowGraph cfg) {
```
msridhar, line 315: Choose a more informative name, like `checkAlwaysCall`.  Also it would be good to write a Javadoc comment with an overview of what this method does.

Also I don't think `entry` needs to be a parameter to the method.  You could just call `cfg.getEntryBlock()` at the start of the method.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+
+    Set<Pair<Long, Set<AssignmentNode>>> visited = new HashSet<>();
+    visited.add(Pair.of(entry.getId(), Collections.emptySet()));
+
+    Deque<Pair<BlockImpl, Set<AssignmentNode>>> worklist = new ArrayDeque<>();
+    worklist.add(Pair.of((BlockImpl) entry, Collections.emptySet()));
```
msridhar, line 321: Why do we use ids in the `visited` list but `BlockImpl`s in the worklist?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+
+    while (!worklist.isEmpty()) {
+
+      Pair<BlockImpl, Set<AssignmentNode>> state = worklist.getLast();
```
msridhar, line 325: You can just do:

```java
      Pair<BlockImpl, Set<AssignmentNode>> state = worklist.removeLast();
```

And then delete the following line.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+        return assignmentNode;
+      }
+    }
+
+    return null;
+  }
+
+  private boolean isVarInPrevDefs(Set<AssignmentNode> defs, LocalVariableNode node) {
+
+    for (AssignmentNode assignmentNode : defs) {
```
msridhar, line 447: You can write this more succinctly and clearly (IMO) using streams:

```java
return defs.stream()
    .map(assign -> ((LocalVariableNode)assign.getTarget()).getElement())
    .anyMatch(elem -> elem.equals(node.getElement()));
```

In any case you should be using `equals()` to compare the elements, not `==` (unless you have a strong reason to do otherwise).

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+
+      for (Node node : nodes) {
+        if (node instanceof AssignmentNode) {
+          Node lhs = ((AssignmentNode) node).getTarget();
+          Node rhs = ((AssignmentNode) node).getExpression();
+
+          if (lhs instanceof LocalVariableNode && hasAlwaysCall(lhs.getType())) {
+
+            // Reassignment to the lhs
+            if (isVarInPrevDefs(newDefs, (LocalVariableNode) lhs)) {
```
msridhar, line 338: Why pass `newDefs` here?  Shouldn't you pass `state.second`?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+              AssignmentNode latestAssignmentNode =
```
msridhar, line 339: Why is this named `latestAssignmentNode`?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+              newDefs.remove(latestAssignmentNode);
+            }
+
+            // If the rhs is ObjectCreationNode, MethodInvocationNode, MethodAccessNode or
+            // LocalVariableNode that exists in the newDefs (Note that if the localVariableNode
+            // exists in the newDefs it means it is assigned to a non-null values), then it adds
+            // the the AssignmentNode to the newDefs.
+            if ((rhs instanceof ObjectCreationNode)
+                || (rhs instanceof MethodInvocationNode)
+                || (rhs instanceof MethodAccessNode)
```
msridhar, line 351: As we discussed we can get rid of the `MethodAccessNode` case

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+                || (rhs instanceof LocalVariableNode
+                    && isVarInPrevDefs(newDefs, (LocalVariableNode) rhs))) {
```
msridhar, line 353: Again I think you should be passing `state.second` here?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+              newDefs.add((AssignmentNode) node);
+            }
+
+            // Ownership Transfer
+            if (rhs instanceof LocalVariableNode
```
msridhar, line 358: I would maybe handle this ownership transfer entirely within this `if`, rather than having the addition of the new assignment node in the previous `if`, for clarity

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+                && isVarInPrevDefs(newDefs, (LocalVariableNode) rhs)) {
```
msridhar, line 359: Again `state.second`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+              newDefs.remove(getAssignmentNodeOfVar(newDefs, (LocalVariableNode) rhs));
+            }
+          }
+        }
+
+        // Remove the returned localVariableNode from newDefs.
+        if (node instanceof ReturnNode) {
+          Node result = ((ReturnNode) node).getResult();
+          if (result instanceof LocalVariableNode
+              && isVarInPrevDefs(newDefs, (LocalVariableNode) result)) {
```
msridhar, line 369: Again `state.second`?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+      for (BlockImpl succ : getSuccessors(state.first)) {
+        Set<AssignmentNode> toRemove = new HashSet<>();
+
+        // There is nothing needs to be checked
+        if (newDefs.size() == 0) {
+          propagate(Pair.of(succ, newDefs), visited, worklist);
+          continue;
+        }
+
+        CFStore succRegularStore = getAnalysis().getInput(succ).getRegularStore();
```
msridhar, line 384: I would call this `succStoreBefore` for clarity

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+
+    for (AssignmentNode assignmentNode : defs) {
+      LocalVariableNode lhs = (LocalVariableNode) assignmentNode.getTarget();
+
+      if (lhs.getElement() == localVarNode.getElement()) {
+        return assignmentNode;
+      }
+    }
+
+    return null;
```
msridhar, line 442: Here I think you should throw an exception rather than just returning `null`.  We only expect this method to be called when a relevant assignment is in the `defs` set

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+    LocalVariableNode lhs = (LocalVariableNode) lastAssignmentNode.getTarget();
+    CFValue lhsCFValue = store.getValue(lhs);
+    String alwaysCallValue = getAlwaysCallValue(lhs.getElement());
+
+    boolean report = true;
+
+    if (lhsCFValue != null) { // When store contains the lhs
+
+      for (AnnotationMirror annotationMirror : lhsCFValue.getAnnotations()) {
+        if (getValueOfAnnotationWithStringArgument(annotationMirror).contains(alwaysCallValue)) {
```
msridhar, line 549: Shouldn't we check if it is a `@CalledMethods` annotation before this check?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -547,6 +844,11 @@ private boolean hasAnnotation(Element element, String annotName) {
     return null;
   }
 
+  public CFAnalysis getAnalysis() {
```
msridhar, line 847: Remove this method; it is no longer needed

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -49,8 +65,27 @@ public Void visitAnnotation(final AnnotationTree node, final Void p) {
...
+        if (!currentCalledMethods.contains(alwaysCallAnnoVal)) {
+          String error = " " + alwaysCallAnnoVal + " has not been called";
+          checker.report(node, new DiagMessage(Diagnostic.Kind.ERROR, "missing.alwayscall", error));
+        }
+      }
+    }
+
     if (checker.getBooleanOption(ObjectConstructionChecker.COUNT_FRAMEWORK_BUILD_CALLS)) {
       ExecutableElement element = TreeUtils.elementFromUse(node);
+
```
msridhar, line 88: Remove this change

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -61,6 +96,81 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
     return super.visitMethodInvocation(node, p);
   }
 
+  @Override
+  public Void visitNewClass(NewClassTree node, Void p) {
+
+    if (!isAssignedToLocal(this.getCurrentPath())) {
```
msridhar, line 102: Please create a method that includes the shared logic from `visitNewClass` and `visitMethodInvocation` and call it in both places, to remove duplication

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/framework/AutoValueSupport.java
```diff
@@ -74,7 +74,8 @@ public boolean isBuilderBuildMethod(ExecutableElement element) {
     if (FrameworkSupportUtils.hasAnnotation(enclosingElement, AutoValue.Builder.class)) {
       assert FrameworkSupportUtils.hasAnnotation(nextEnclosingElement, AutoValue.class)
           : "class " + nextEnclosingElement.getSimpleName() + " is missing @AutoValue annotation";
-      // it is a build method if it returns the type with the @AutoValue annotation
+
```
msridhar, line 77: Remove this change

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: I think it should be newDefs, consider the following test case:
void ownershipTransfer(){
        Foo f1 = new Foo();
        Foo f2 = f1;
   }
the state is Pair(RegularBlock([f1 = new Foo(), f2, f1, f2 = f1,]), []), so state.second is empty list, but newDefs contains f1 = new Foo() after the first node. Then, the result of isVarInPrevDefs(state.second, rhs) in the following check is different from isVarInPrevDefs(newDefs, rhs):
if ((rhs instanceof ObjectCreationNode)
                || (rhs instanceof MethodInvocationNode)
                || (rhs instanceof LocalVariableNode
                    && isVarInPrevDefs(newDefs, (LocalVariableNode) rhs))){
          newDefs.add((AssignmentNode) node);
}

Nargeshdb: I think it should be newDefs, consider the following example:
    void ownershipTransfer(){
        Foo f1 = new Foo();
        Foo f2 = new Foo();
        f2 = f1;
        f2.a();
    }
The state is Pair(RegularBlock([f2 = new Foo(), f2, f1, f2 = f1, f2]), [f1 = new Foo()]) when we are checking f2 = f1, so state.second does not contain f2 = new Foo(), but newDefs contains {f1 = new Foo(), f2 = new Foo()}. Thus, if we use state.second instead of newDefs, then isVarInPrevDefs(state.second, (LocalVariableNode) lhs) returns false and it won't report error for Foo f2 = new Foo();

msridhar: Ok, I see.  My pseudocode assumed a single statement per block.  Here, I think we should change the code and have `prevDefs` and `newDefs` available in the loop over nodes in a block.  At the beginning of each loop iteration, `newDefs` should be a new Set containing `prevDefs`, and at the end of each loop iteration, we should have `prevDefs = newDefs`.  Then, all calls to `isVarInPrevDefs` should pass `prevDefs`.  Does that make sense?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+              && isVarInPrevDefs(newDefs, (LocalVariableNode) result)) {
+            newDefs.remove(getAssignmentNodeOfVar(newDefs, (LocalVariableNode) result));
+          }
+        }
+      }
+
+      for (BlockImpl succ : getSuccessors(state.first)) {
+        Set<AssignmentNode> toRemove = new HashSet<>();
+
+        // There is nothing needs to be checked
```
msridhar, line 378: This check is unnecessary.  If `newDefs` is empty then the next loop will be skipped and the right thing will happen.  So we should get rid of it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+        }
+
+        CFStore succRegularStore = getAnalysis().getInput(succ).getRegularStore();
+
+        for (AssignmentNode assignmentNode : newDefs) {
+
+          // If the successor block is the exit block or if the set of the pair of block id and set
+          // of defs is repeated in a loop or if the variable is going out of scope
+          if (succ instanceof SpecialBlockImpl
+              || visited.contains(Pair.of(succ.getId(), newDefs))
```
msridhar, line 391: What happens if you remove this case?  I don't understand why it is needed

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +306,276 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+                  assignmentNode,
+                  storeAfter,
+                  (last instanceof AssignmentNode) ? getAnnotatedType(last.getTree()) : null);
+            }
+          }
+
+          if (succRegularStore.getValue((LocalVariableNode) (assignmentNode.getTarget())) == null) {
+            // If the lhs of the assignmentNode isn't in the successor's input store, it means the
+            // variable is going out of scope
+            toRemove.add(assignmentNode);
```
msridhar, line 412: Can't this call go inside the previous `if` block?  Why do we need to check this condition again?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: Why should we change newDefs and use prevDefs? do you have any example to show the difference?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: Because it's possible to have the same pair<block, set<defs>>, but we should check both of them. Consider the following example:
    void overWrittingVarInLoop() {
        // :: error: missing.alwayscall
        Foo f = new Foo();
        while (true) {
            // :: error: missing.alwayscall
            f = new Foo();
        }
    }
If we use BlockImpl in the visited then it cannot find the second error inside the loop. The reason that I kept BlockImpl in the worklist is that I couldn't retrieve blocks with just blockIds.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: Then it can't find errors inside the loops properly:
void overWrittingVarInLoop() {
        // :: error: missing.alwayscall
        Foo f = new Foo();
        while (true) {
            // :: error: missing.alwayscall
            f = new Foo();
        }
    }

The other example is:
    void loopWithNestedBranches(boolean b) {
        Foo f = null;
        while (true) {
            if (b) {
                // :: error: missing.alwayscall
                f = new Foo();
            } else {
                // :: error: missing.alwayscall
                f = new Foo();
            }
        }
    }

It doesn't report an error for the if branch! But it reports an error for the then branch because it's checked first and then the if branch overwrites it!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
msridhar: I think now that the code is correct as-is.  But let's rename the method `isVarInPrevDefs` to just `isVarInDefs` to avoid confusion.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,253 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+  @Override
+  public void postAnalyze(ControlFlowGraph cfg) {
+    checkAlwaysCall(cfg);
+    super.postAnalyze(cfg);
+  }
+
+  private void checkAlwaysCall(ControlFlowGraph cfg) {
+
+    Set<Pair<BlockImpl, Set<Tree>>> visited = new HashSet<>();
+    visited.add(Pair.of((BlockImpl) cfg.getEntryBlock(), Collections.emptySet()));
```
msridhar, line 316: For clarity let's store `Pair.of((BlockImpl) cfg.getEntryBlock(), Collections.emptySet())` in a local variable and re-use it for `visited` and `worklist` initialization

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,253 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+              newDefs.remove(Pair.of((LocalVariableNode) lhs, latestAssignmentPair));
+            }
+
+            // If the rhs is ObjectCreationNode, MethodInvocationNode, or
+            // LocalVariableNode that exists in the newDefs (Note that if the localVariableNode
+            // exists in the newDefs it means it is assigned to a non-null values), then it adds
+            // the the AssignmentNode to the newDefs.
+            if ((rhs instanceof ObjectCreationNode)
+                || (rhs instanceof MethodInvocationNode)
+                || (rhs instanceof LocalVariableNode
```
msridhar, line 348: Can we remove this condition here, and fully handle ownership transfer in the condition below (addition and deletion from newDefs)?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,253 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
 
+  @Override
+  public void postAnalyze(ControlFlowGraph cfg) {
+    checkAlwaysCall(cfg);
+    super.postAnalyze(cfg);
+  }
+
+  private void checkAlwaysCall(ControlFlowGraph cfg) {
+
+    Set<Pair<BlockImpl, Set<Tree>>> visited = new HashSet<>();
```
msridhar, line 315: Let's keep the type of `visited` and `worklist` the same.  So `visited` should be `Set<Pair<BlockImpl, Set<Pair<LocalVariableNode, Tree>>>>`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,253 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+   */
+  private void propagate(
+      Pair<BlockImpl, Set<Pair<LocalVariableNode, Tree>>> state,
+      Set<Pair<BlockImpl, Set<Tree>>> visited,
+      Deque<Pair<BlockImpl, Set<Pair<LocalVariableNode, Tree>>>> worklist) {
+
+    if (!visited.contains(
+        Pair.of(
+            state.first,
+            state.second.stream().map(item -> item.second).collect(Collectors.toSet())))) {
```
msridhar, line 491: This will be simpler once we make the types of `visited` and `worklist` the same

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -49,6 +63,23 @@ public Void visitAnnotation(final AnnotationTree node, final Void p) {
   @Override
   public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
 
+    if (!isAssignedToLocal(this.getCurrentPath()) && !atypeFactory.returnsThis(node)) {
```
msridhar, line 66: We can leave this logic for now, but it will likely be much simpler or even gone in the future once we can track types of particular expressions in the dataflow analysis

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -61,6 +92,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
     return super.visitMethodInvocation(node, p);
   }
 
+  @Override
+  public Void visitNewClass(NewClassTree node, Void p) {
+    if (!isAssignedToLocal(this.getCurrentPath())) {
+      TypeMirror type = TreeUtils.typeOf(node);
+      if (hasAlwaysCall(type)) {
```
msridhar, line 99: 👍

object-construction-checker/tests/basic/ACExceptionalExitPointTest.java
```diff
@@ -0,0 +1,38 @@
+import org.checkerframework.checker.objectconstruction.qual.*;
+import org.checkerframework.common.returnsreceiver.qual.*;
+
+class ACExceptionalExitPointTest {
```
msridhar, line 4: Can we move these `AC*` tests to a folder `object-construction-checker/tests/alwayscall` and then run them in a separate test class?

object-construction-checker/tests/file/ACFileTest.java
```diff
@@ -0,0 +1,5 @@
+
+
+public class FileACTest {
```
msridhar, line 3: Can we keep all the file- and socket-related changes out of this PR?  This includes the size 0 `.astub` files; not sure how they even got added here.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: Yes, we can! Thanks!

object-construction-checker/tests/file/ACFileTest.java
Nargeshdb: Sorry! I added this by mistake!

msridhar:
@kelloggm I think this now ready for you to take a look

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+        break;
+
+      case EXCEPTION_BLOCK:
+        blockNodes.add(((ExceptionBlockImpl) block).getNode());
+    }
+    return blockNodes;
+  }
+
+  /**
+   * It updates visited and worklist if the input state has not been visited yet
```
kelloggm, line 477: Nit: the style of this comment is a bit confusing. I would rewrite it as:

`Updates {@code vistited} and {@code worklist} if the input {@code state} has not been visited yet.`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+      Set<Pair<BlockImpl, Set<Pair<LocalVariableNode, Tree>>>> visited,
+      Deque<Pair<BlockImpl, Set<Pair<LocalVariableNode, Tree>>>> worklist) {
+
+    if (!visited.contains(state)) {
+      visited.add(state);
+      worklist.add(state);
+    }
+  }
+
+  private String getAlwaysCallValue(Element element) {
```
kelloggm, line 494: return type should be `@Nullable String`, and there should be a simple Javadoc comment summarizing the effect of the method.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+    TypeMirror type = element.asType();
+    TypeElement eType = TypesUtils.getTypeElement(type);
+    AnnotationMirror alwaysCallAnnotation = getDeclAnnotation(eType, AlwaysCall.class);
+
+    return (alwaysCallAnnotation != null)
+        ? AnnotationUtils.getElementValue(alwaysCallAnnotation, "value", String.class, false)
+        : null;
+  }
+
+  private void reportAlwaysCallErrors(
```
kelloggm, line 505: This method actually does significantly more than just report always call errors. I would change the name to something like "checkAlwaysCall" or at least write Javadoc to that effect (it should have Javadoc anyway).

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+    CFValue lhsCFValue = store.getValue(assign.first);
+    String alwaysCallValue = getAlwaysCallValue(assign.first.getElement());
+
+    boolean report = true;
+
+    if (lhsCFValue != null) { // When store contains the lhs
+
+      for (AnnotationMirror annotationMirror : lhsCFValue.getAnnotations()) {
+        if (AnnotationUtils.areSameByClass(annotationMirror, CalledMethods.class)
+            && getValueOfAnnotationWithStringArgument(annotationMirror).contains(alwaysCallValue)) {
```
kelloggm, line 519: Rather than checking this directly, can you construct the appropriate `@CalledMethods` annotation that corresponds to the `@AlwaysCall` annotation and then do a subtyping check? That would allow e.g. `@CalledMethodsPredicate` annotations that are subtypes of `@CalledMethods` annotations (using one of the extension points that I added in the Accumulation predicate PR) to also be allowed.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+      AnnotationMirror annotationMirror = annotatedTypeMirror.getAnnotation(CalledMethods.class);
+
+      if (annotationMirror != null
+          && getValueOfAnnotationWithStringArgument(annotationMirror).contains(alwaysCallValue)) {
+        report = false;
+      }
+    }
+
+    if (report) {
+      checker.report(
```
kelloggm, line 538: use `checker.reportError`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -49,6 +63,23 @@ public Void visitAnnotation(final AnnotationTree node, final Void p) {
...
+      if (hasAlwaysCall(returnType)) {
+        TypeElement eType = TypesUtils.getTypeElement(returnType);
+        AnnotationMirror alwaysCallAnno = atypeFactory.getDeclAnnotation(eType, AlwaysCall.class);
+
+        String alwaysCallAnnoVal =
+            AnnotationUtils.getElementValue(alwaysCallAnno, "value", String.class, false);
+        List<String> currentCalledMethods = getCalledMethodAnnotation(node);
+
+        if (!currentCalledMethods.contains(alwaysCallAnnoVal)) {
+          checker.report(node, new DiagMessage(Diagnostic.Kind.ERROR, "missing.alwayscall", ""));
```
kelloggm, line 78: Use `checker.reportError`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -49,6 +63,23 @@ public Void visitAnnotation(final AnnotationTree node, final Void p) {
...
+
+      if (hasAlwaysCall(returnType)) {
+        TypeElement eType = TypesUtils.getTypeElement(returnType);
+        AnnotationMirror alwaysCallAnno = atypeFactory.getDeclAnnotation(eType, AlwaysCall.class);
+
+        String alwaysCallAnnoVal =
+            AnnotationUtils.getElementValue(alwaysCallAnno, "value", String.class, false);
+        List<String> currentCalledMethods = getCalledMethodAnnotation(node);
+
+        if (!currentCalledMethods.contains(alwaysCallAnnoVal)) {
```
kelloggm, line 77: This should also probably use `isSubtype` from the type factory rather than implement the check directly.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -61,6 +92,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
     return super.visitMethodInvocation(node, p);
   }
 
+  @Override
+  public Void visitNewClass(NewClassTree node, Void p) {
+    if (!isAssignedToLocal(this.getCurrentPath())) {
+      TypeMirror type = TreeUtils.typeOf(node);
+      if (hasAlwaysCall(type)) {
+        checker.report(node, new DiagMessage(Diagnostic.Kind.ERROR, "missing.alwayscall", " "));
```
kelloggm, line 100: use `checker.reportError`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -61,6 +92,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+      }
+    }
+    return super.visitNewClass(node, p);
+  }
+
+  private List<String> getCalledMethodAnnotation(MethodInvocationTree node) {
```
kelloggm, line 106: Rather than this method, use the type factory's `getAnnotatedType` method and then `isSubtype`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
```diff
@@ -61,6 +92,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+          // No point in going on.
+          return false;
+        }
+        // Otherwise use the context of the ConditionalExpressionTree.
+        return isAssignedToLocal(parentPath);
+      case ASSIGNMENT: // check if the left hand is a local variable
+        // TODO! not sure about this part!
+        final JCTree.JCExpression lhs = ((JCTree.JCAssign) parent).lhs;
+        if (!(lhs instanceof JCTree.JCFieldAccess)) {
+          return (((JCTree.JCIdent) lhs).sym.getKind() == LOCAL_VARIABLE);
```
kelloggm, line 143: Rather than casting to `JCIdent` no matter what if this isn't a field access, I think we should be defensive and check whether the class, and throw a `BugInCF` exception if it's not `JCIdent`

object-construction-checker/tests/alwayscall/ACRegularExitPointTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+
+    void loopWithNestedBranches(boolean b) {
+        Foo f = null;
+        while (true) {
+            if (b) {
+                // :: error: missing.alwayscall
+                f = new Foo();
+            } else {
+                //TODO
```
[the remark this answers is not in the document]
Nargeshdb: It relates to the second open problem in the design doc. If we have f.a() inside the else branch, the checker reports an error for "f = new Foo()" inside the else branch. It will be fixed in the next pull request.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+            // the the AssignmentNode to the newDefs.
+            if ((rhs instanceof ObjectCreationNode) || (rhs instanceof MethodInvocationNode)) {
+              newDefs.add(Pair.of((LocalVariableNode) lhs, node.getTree()));
+            }
+
+            // Ownership Transfer
+            if (rhs instanceof LocalVariableNode && isVarInDefs(newDefs, (LocalVariableNode) rhs)) {
+              // If the rhs is a LocalVariableNode that exists in the newDefs (Note that if a
+              // localVariableNode exists in the newDefs it means it is assigned to a non-null
+              // values), then it adds the localVariableNode to the newDefs
```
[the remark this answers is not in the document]
msridhar: Yes, this is referring to null literals.  We should clarify

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionVisitor.java
Nargeshdb: Should I make a dummy CalledMethod type out of AlwaysCall to use isSubtype?
Could you please clarify why you think it's the cleaner way? 
@kelloggm

kelloggm: > Should I make a dummy CalledMethod type out of AlwaysCall to use isSubtype

Yes, that's what I had in mind.

> Could you please clarify why you think it's the cleaner way?

Sure. There are two reasons:
1. it avoids duplicating code, following the DRY ("don't repeat yourself) principle
2. it will enable these checks to take advantage of improvements in the `isSubtype` code. For example, if the type is an `@CalledMethodsPredicate` annotation and the checker was using an SMT solver to allow precise subtyping between those annotations and `@CalledMethods` annotations, then using `isSubtype` would allow the check here to succeed. If you re-implement, as you've done here, you wouldn't be able to take advantage of such improvements.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +303,248 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+      ConditionalBlock ccur = (ConditionalBlock) cur;
+
+      successorBlock.add((BlockImpl) ccur.getThenSuccessor());
+      successorBlock.add((BlockImpl) ccur.getElseSuccessor());
+
+    } else {
+      assert cur instanceof SingleSuccessorBlock;
+
+      Block b = ((SingleSuccessorBlock) cur).getSuccessor();
+      if (b != null) {
```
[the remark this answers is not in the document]
kelloggm: Can you address this comment?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,313 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+
+    CFValue lhsCFValue = store.getValue(assign.first);
+    String alwaysCallValue = getAlwaysCallValue(assign.first.getElement());
+
+    boolean report = true;
+
+    if (lhsCFValue != null) { // When store contains the lhs
+      AnnotationMirror dummyCMAnno = createCalledMethods(alwaysCallValue);
+      AnnotatedTypeMirror annoType = getAnnotatedType(assign.first.getTree());
+      AnnotationMirror CMAnno = annoType.getAnnotationInHierarchy(TOP);
```
kelloggm, line 575: nit: the name of this variable shouldn't start with a capital letter. The standard Java convention is to use camel-case for all local variables, so here I would name this `cmAnno`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,313 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+      QualifierHierarchy qualifierHierarchy =
+          createQualifierHierarchy(new MultiGraphQualifierHierarchy.MultiGraphFactory(this));
```
kelloggm, line 577: It shouldn't be necessary to create a new qualifier hierarchy - the type factory should already have one. You can use `AnnotatedTypeFactory#getQualifierHierarchy` instead.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,313 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+  }
+
+  private boolean hasAlwaysCall(TypeMirror type) {
+    TypeElement eType = TypesUtils.getTypeElement(type);
+    return ((eType != null) && (getDeclAnnotation(eType, AlwaysCall.class) != null));
+  }
+
+  private class BlockWithLocals {
+    public BlockImpl block;
+    public Set<Pair<LocalVariableNode, Tree>> localSet;
```
kelloggm, line 606: We might also consider giving `Pair<LocalVariableNode, Tree>` a name - that structure is used extensively above, and also here.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,313 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+
+    if (report) {
+      checker.reportError(assign.second, "missing.alwayscall", alwaysCallValue);
+    }
+  }
+
+  private boolean hasAlwaysCall(TypeMirror type) {
+    TypeElement eType = TypesUtils.getTypeElement(type);
+    return ((eType != null) && (getDeclAnnotation(eType, AlwaysCall.class) != null));
+  }
```
kelloggm, line 602: This method appears both here and in `ObjectConstructionVisitor`. Can you consolidate them into a single method and make it package-private.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -279,6 +304,313 @@ private static String getExactStringValue(Tree tree, ValueAnnotatedTypeFactory f
...
+      AnnotatedTypeMirror annotatedTypeMirror) {
+
+    CFValue lhsCFValue = store.getValue(assign.first);
+    String alwaysCallValue = getAlwaysCallValue(assign.first.getElement());
+
+    boolean report = true;
+
+    if (lhsCFValue != null) { // When store contains the lhs
+      AnnotationMirror dummyCMAnno = createCalledMethods(alwaysCallValue);
+      AnnotatedTypeMirror annoType = getAnnotatedType(assign.first.getTree());
```
msridhar, line 574: I'm a bit confused.  We get the value of `assign.first` from the store but we never use it other than a null check.  Shouldn't we be getting the type of `assign.first` from `lhsCFValue` here?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -91,6 +132,7 @@
    */
   public ObjectConstructionAnnotatedTypeFactory(final BaseTypeChecker checker) {
     super(checker);
+    stubTypes = new StubTypes(this);
```
msridhar, line 135: @Nargeshdb you are not actually using this, right?  So we can delete?

Nargeshdb: Yes, I am not using this! It's deleted!

PR #207, opened 2020-10-16
AlwaysCall declaration annotation -> MustCall type annotation

kelloggm:
This PR implements a `MustCall` typechecker, which replaces the `@AlwaysCall` declaration annotation used previously. All the tests pass.

There are some differences between `@AlwaysCall` and `@MustCall`:
* `@MustCall` takes an array of strings as an argument, rather than a single string. This means that the checker can enforce stronger specifications, such as `@MustCall({"read", "close"})`.
* `@MustCall` is *not* inheritable, because it is a type annotation. Rather, errors are reported in subclasses that don't conform to the declared type of the super class. This leads to the only new warning in the tests (`super.invocation.invalid` on a custom class defined as `@MustCall({})` whose superclass is `@MustCall({"a"})`, which I judge to be a true positive). It also leads to a lot of changes in the stub files: it is no longer sufficient to annotate just the `Closeable` interface. It was a pain to get all of these in testing, and it will be a pain as we deploy. It might be worth looking into making the checker automatically read type annotations off of superclasses and superinterfaces. I have some code that's a start on that, but it's not working yet and I didn't want to delay this PR; expect a follow up.

Note that to update our larger benchmarks, you'll need to add a compile-time dependency on `net.sridharan.objectconstruction:must-call-qual:0.1.10-SNAPSHOT`, because the must call qualifiers are in a separate project in the same manner as the old Returns Receiver qualifiers.


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Default constructor matching super. Should be called automatically.
+   *
+   * @param checker the checker associated with this type factory
+   */
+  public MustCallAnnotatedTypeFactory(final BaseTypeChecker checker) {
+    super(checker);
+    TOP = AnnotationBuilder.fromClass(elements, MustCallTop.class);
+    BOTTOM = createMustCall();
```
msridhar, line 41: Why don't we need `@MustCallBottom`?  We had `@CalledMethodsBottom`, distinguished from `@CalledMethods({})`.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Set;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.ExecutableElement;
+import org.checkerframework.common.basetype.BaseTypeChecker;
+import org.checkerframework.common.basetype.BaseTypeVisitor;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
+
+/**
+ * The visitor for the MustCall checker. This visitor is similar to BaseTypeVisitor, but overrides
+ * methods that don't work well with the MustCall type hierarchy because it uses the bottom type as
```
msridhar, line 14: Would this be solved by having a `@MustCallBottom`?

msridhar (commented):
Overall looks great!  The lack of inheritance of `@MustCall` is a bit concerning, as we could miss library classes with `@MustCall` obligations.  We should address it but I agree it doesn't need to be in this PR.  Just a couple comments (which are really the same comment)

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -386,9 +390,13 @@ private void alwaysCallTraverse(ControlFlowGraph cfg) {
...
               && isVarInDefs(newDefs, (LocalVariableNode) receiver)) {
-            if (getAlwaysCallValue(((LocalVariableNode) receiver).getElement())
-                .equals(method.getSimpleName().toString())) {
-              // If the method called on the receiver is the same as receiver's @AlwaysCall value,
+            List<String> mustCallVal = getMustCallValue(((LocalVariableNode) receiver).getTree());
+            // TODO: I think this will cause false positives if there is more than one value in the
+            // @MustCall annotation
+
+            if (mustCallVal.size() == 1
+                && mustCallVal.get(0).equals(method.getSimpleName().toString())) {
```
Nargeshdb, line 398: I think we should remove the assignment tree from the newDefs if mustCallVal.size() is greater than 1 and the node is the last required method that needs to be called. But, it's not important for this pull request I can add this check later.

Nargeshdb (commented):
It looks great to me too. Thanks, Martin.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
kelloggm: `@CalledMethodsTop` is equivalent to `@CalledMethods({})`. We have `@CalledMethodsBottom` for the same reason that `@MustCallTop` is required here: there are a potentially-infinite number of `@MustCall(...)` annotations, and they need to have a single supertype; similarly, there are a potentially-infinite number of `@CalledMethods(...)` types, and they need a single subtype. The difference is the direction of the type hierarchy: `@MustCall({})` is the bottom type, but `@CalledMethods({})` is the top type.

We actually don't need `@CalledMethodsTop`: it was included so that predicates had a sensible LUB with `@CalledMethods` annotations, but even that isn't needed. The Called Methods Checker in the CF master branch doesn't have an `@CalledMethodsTop` annotation and instead just uses `@CalledMethods({})` as the top type.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
kelloggm: No, because the problem is related to any non-top default. I'll clarify this documentation.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
msridhar: That makes sense, thanks!

msridhar:
Another question here.  Say we have code like:
```java
Object o = new FileInputStream(...);
```
With this change, I think we should get an error for this code where we didn't before (since we will now infer the type of `o` to be `@MustCall("close") Object`.  Do we get an error for this example now?

kelloggm:
> Another question here. Say we have code like:
> 
> ```java
> Object o = new FileInputStream(...);
> ```
> 
> With this change, I think we should get an error for this code where we didn't before (since we will now infer the type of `o` to be `@MustCall("close") Object`. Do we get an error for this example now?

We should. I've add a passing unit test case to confirm.

kelloggm:
@mernst, as we discussed, this PR should be ready for you to review. You should focus on the new code (i.e. the code in the `must-call-checker` and `must-call-qual` directories); we'll ask you to review the rest (i.e. the integration with the existing worklist algorithm) later when you review the full worklist algorithm.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+   * concerned with annotation arguments (which must be literals, and therefore won't have (or be
+   * able to fulfill) must-call obligations).
+   */
+  @Override
+  public Void visitAnnotation(AnnotationTree node, Void p) {
+    return null;
+  }
+
+  /**
+   * Only check type arguments that are bottom, to avoid false positives.
```
mernst, line 91: This choice seems likely to be unsound.  If it is intentionally unsound, please note that.  If it is sound, please give an argument why.
I tried to write some tests, but they crashed the checker.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+   * Sockets does in fact have a responsibility to close each of those sockets.
+   */
+  protected void checkTypeArguments(
+      Tree toptree,
+      List<? extends AnnotatedTypeParameterBounds> paramBounds,
+      List<? extends AnnotatedTypeMirror> typeargs,
+      List<? extends Tree> typeargTrees,
+      Name typeOrMethodName,
+      List<?> paramNames) {
+    List<Integer> indicesToRemove = new ArrayList<>();
```
mernst, line 106: There is no need for this list.  You can combine the two `for` loops.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+        toptree,
+        paramBoundsWithoutNonDefault,
+        typeargsWithoutNonDefault,
+        typeargTrees,
+        typeOrMethodName,
+        paramNames);
+  }
+
+  /**
+   * Does a deep scan of the annotated type for any non-default (that is, non-bottom) annotation.
```
mernst, line 133: Non-default and non-bottom are different.  The default is `@CalledMethods({})`.  Bottom is `@CalledMethods({a, b, c, d, ...})`.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
```diff
@@ -0,0 +1,46 @@
...
+    return handleStringConcatenation(super.visitStringConcatenate(n, input));
+  }
+
+  @Override
+  public TransferResult<CFValue, CFStore> visitStringConcatenateAssignment(
+      StringConcatenateAssignmentNode n, TransferInput<CFValue, CFStore> in) {
+    return handleStringConcatenation(super.visitStringConcatenateAssignment(n, in));
+  }
+
+  private TransferResult<CFValue, CFStore> handleStringConcatenation(
```
mernst, line 39: Please document to aid in understanding.  This sets all string concatenations to `@CalledMethods({})`.
Why is that necessary in the transfer function?  That is, why cannot the AnnotatedTypeFactory do this?

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,181 @@
...
```
mernst, line 52: According to the [type hierarchy in `MustCall.java`](https://github.com/kelloggm/object-construction-checker/pull/207/files#diff-2718c67b17c6924728dd0c221267cf949ab92ae7cf0f71ea33c3320bb8128109R19),
`@MustCall({"m1"}) :> @MustCall({"m1", "m2"})`.
That means that the variable `BOTTOM` is not the bottom type.  Maybe you should rename it to `DEFAULT`, and have a new variable `BOTTOM` that really is bottom.

mernst (changes requested):
My biggest concern with this pull request is its treatment of the bottom type:  there isn't one, and the variable named `BOTTOM` is not at the bottom of the type hierarchy.  This also means that methods like `greatestLowerBound` are incorrect.  I suggest adding a new `MustCallBottom.java` annotation to clean up these problems.  I don't see the need for a distinct top annotation, and the current one's documentation is a description of a bottom type.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
msridhar: I think the [linked type hierarchy comment](https://github.com/kelloggm/object-construction-checker/pull/207/files#diff-2718c67b17c6924728dd0c221267cf949ab92ae7cf0f71ea33c3320bb8128109R19) is incorrect, which is causing confusion.  With `@MustCall({"m1"}) :> @MustCall({"m1", "m2"})`, the subtype could _add_ must-call obligations, which then (unsoundly) wouldn't be checked if a value is assigned to a supertype variable.  Instead, we want `@MustCall({"m1"}) <: @MustCall({"m1", "m2"})`, so subtyping requires a _subset_ of must-call obligations.  I think @kelloggm implemented this correctly (so `@MustCall({})` is indeed the bottom type), but the linked comment is wrong.

kelloggm: Manu is right - the comment is wrong (somehow I always seem to make typos when I use subtyping symbols...). `@MustCall({})` is the bottom type (and the default) in the hierarchy.

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCallTop.java
```diff
@@ -0,0 +1,16 @@
+package org.checkerframework.checker.mustcall.qual;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The top qualifier in the Must Call type hierarchy. It represents a type that must have all of its
```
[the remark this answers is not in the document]
kelloggm: Most of this comment is irrelevant because of the incorrect documentation of the type hierarchy. However, I agree that `*Top` isn't a very useful name. In this type system's case, though, `MustCallUnknown` wouldn't be right - it's more like "must-call unknowable/unsatisfiable". I therefore decided to rename it `MustCallAll` and add documentation that indicated that it can never be satisfied (because the "all" it represents really means "all possible methods", not just defined on the class).

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+ */
+public class MustCallVisitor extends BaseTypeVisitor<MustCallAnnotatedTypeFactory> {
+
+  /** @param checker the type-checker associated with this visitor */
+  public MustCallVisitor(BaseTypeChecker checker) {
+    super(checker);
+  }
+
+  /**
+   * Typically issues a warning if the result type of the constructor is not top. This is not a
```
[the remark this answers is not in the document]
kelloggm: This paragraph is a defense for why it makes sense for this method not to issue a warning. I'm inclined to keep it as-is, with an additional first paragraph that states that this implementation of the method does not issue a warning.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
kelloggm: This choice was unsound. I changed the defaulting rules for generics to make this unnecessary.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
kelloggm: > That is, why cannot the AnnotatedTypeFactory do this?

It could. I borrowed this implementation from the Value Checker, which also has custom handling for string concatenation (e.g. [here](https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java#L590)).

I think doing this in the transfer is more performant than doing it as part of the ATF, but the effect is the same - the type of all String concatenations is bottom. The reason to do so is that implicitly calling `toString()` on an object with a must-call obligation doesn't create a corresponding must-call obligation for the string. Without this code, the below would not pass the OCC's must-call check because the argument to `println` would need to call `close` (which it can't, because it's a string!):

```
void test(@Owning Socket s) {
    System.out.println("this is my socket: " + s);
    s.close();
}
```

Obviously that code does call `close`, so the warning is a false positive.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+   */
+  @Override
+  protected void checkConstructorResult(
+      AnnotatedExecutableType constructorType, ExecutableElement constructorElement) {
+    // Do nothing
+  }
+
+  /**
+   * Change the default for exception parameter lower bounds to bottom (the default), to prevent
+   * false positives. I think it might be a bug in the Checker Framework that these locations are
```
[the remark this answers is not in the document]
kelloggm: https://github.com/typetools/checker-framework/issues/3839

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+   * @return a set containing only the @MustCall({}) annotation
+   */
+  @Override
+  protected Set<? extends AnnotationMirror> getExceptionParameterLowerBoundAnnotations() {
+    return Collections.singleton(atypeFactory.BOTTOM);
+  }
+
+  /**
+   * The Checker Framework's default implementation of this method defers to {@code
+   * #getExceptionParameterLowerBoundAnnotations}. That is a bug; this method should always return
```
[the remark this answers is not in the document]
kelloggm: When creating https://github.com/typetools/checker-framework/issues/3839, I tried to write a test that required this change but couldn't. I'll remove it here.

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCall.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * If an expression has type {@code @MustCall({"m1", "m2"})}, then its value has an obligation to
+ * eventually call methods {@code m1} and {@code m2} before any exit point or de-allocation.
+ *
+ * <p>The subtyping relationship is:
+ *
+ * <pre>{@code @MustCall({"m1"}) <: @MustCall({"m1", "m2"})}</pre>
+ *
+ * <p>For example, an object with an obligation to call only "m1" can be passed to a method that
```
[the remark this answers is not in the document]
kelloggm: This type system is unorthodox, because it is not tracking a fact about the values in the program directly. Rather, it is tracking an obligation for each value that is a superset of what it should do before it is de-allocated.

An object with the type `@MustCall({"m1", "m2"}) Object`, just as `param` in your example, has an obligation to call `m1()` and `m2` before it is de-allocated. This says nothing about what the object might actually contain; it is stating a fact that is otherwise external to the existing type system. In other words, this type system in a sense tracks provenance (like the tainting checker) rather than the possible values that an object might take on (like the value checker).

Maybe the confusion here stems from the fact that this checker doesn't actually enforce this property directly; like the Value Checker, it is intended only to provide information to other checkers (in this case, the must-call analysis embedded in the Object Construction Checker proper). It's worth noting that that analysis *isn't* a type system; rather, it's a dataflow analysis that runs directly on the CFG after both the MustCall checker here and the Called Methods checker. The facts that those two checkers produce allow that dataflow analysis to do its job, which is to ensure that each MustCall obligation has been fulfilled.

Another way to think about this is that `@MustCall({"m1", "m2"}) Object` represents all objects that need to call `m1`, `m2`, both, or neither. It cannot represent an object that needs to call some *other* method. So, for example, `@MustCall({"m1", "m2"}) Object` represents more possible values than `@MustCall({"m1"}) Object` does, because `@MustCall({"m1"}) Object` can only contain an object that needs to call `m1` or one that needs to call nothing - an object that needs to call `m2` (or both `m1` and `m2`) is not permitted. This view makes the need for `@MustCallTop` more apparent: `@MustCallTop` represents all objects, which might need to call any set of their methods.

Hopefully this makes the design of the type system make sense. I'm sure we can find ways to make the documentation both clearly express this and more concise than what I've written above. I've made an attempt to rewrite the definition of the `@MustCall` annotation. Can you take a look at let me know if that makes it more clear?

> Maybe you can recast the annotation as `@WillCall`.

I don't think that's quite right, either, but I don't have a better suggestion. `@MustCall` makes intuitive sense to me (and I think it did to you, too, when we discussed the idea on the phone a few weeks ago), but I agree that it doesn't quite capture the notion of what I've described in the preceding paragraphs.

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCall.java
```diff
@@ -0,0 +1,40 @@
...
+ * If an expression has type {@code @MustCall({"m1", "m2"})}, then its value might have an
+ * obligation to eventually call methods {@code m1}, {@code m2}, both, or neither before any exit
+ * point or de-allocation. The value cannot have an obligation to call any methods that are not
+ * named in the annotation.
+ *
+ * <p>The subtyping relationship is:
+ *
+ * <pre>{@code @MustCall({"m1"}) <: @MustCall({"m1", "m2"})}</pre>
+ *
+ * <p>For example, an object with an obligation to call only "m1" can be passed to a method whose
```
mernst, line 23: "an obligation to call only m1" does not refer to anything that this type system can represent.  Should "only" be "at most"?  Likewise on the next line, should "both" be "at most both"?

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCall.java
```diff
@@ -0,0 +1,40 @@
...
+ * argument has an obligation to call both "m1" and "m2", because if both "m1" and "m2" are called,
```
mernst, line 24: The "if both ..." text is different than the wording in [your comment](https://github.com/kelloggm/object-construction-checker/pull/207#discussion_r513117386).  Can you make them consistent?
My concern is that the comment about "if both m1 and m2 are called" does not follow, because a `@MustCall(m1, m2)` might have neither m1 nor m2 called on it.

As a general rule, it's usually better to just update the pull request, rather than writing an explanation in the pull request.  People reading the code won't have access to the pull request conversation, and the code should stand on its own.

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCallAll.java
```diff
@@ -0,0 +1,17 @@
+package org.checkerframework.checker.mustcall.qual;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The top qualifier in the Must Call type hierarchy. It represents a type that must have all of its
```
mernst, line 10: This comment is inconsistent with the new text in `MustCall.java`.  Please make them consistent.

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCall.java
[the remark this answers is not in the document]
mernst: > Another way to think about this is that `@MustCall({"m1", "m2"}) Object` represents all objects that need to call `m1`, `m2`, both, or neither. It cannot represent an object that needs to call some _other_ method.

This is a quite different definition than was previously documented.  Thanks for the update; that helps.

Maybe a name like `MightMustCall` or `MustCallAtMost` would better convey this new semantics.


Your arguments about being unorthodox or not enforcing or not being a type system don't make sense to me.  Even if this were only a dataflow analysis, it would still need a clear definition of the dataflow lattice, in which lower lattice elements represent fewer values.  Hopefully we are converging on that, so that is good news.

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCall.java
kelloggm: It is awkward to say "at most" every time one mentions "obj must call foo". I've added a note as part of the new documentation in the README.md file that should make it clear that any uses of the shorthand "obj must call foo" actually mean "obj must call at most foo".

must-call-qual/src/main/java/org/checkerframework/checker/mustcall/qual/MustCall.java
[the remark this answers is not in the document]
kelloggm: > This is a quite different definition than was previously documented. Thanks for the update; that helps.

Thanks for helping me make the documentation match what the code actually does!

> Maybe a name like MightMustCall or MustCallAtMost would better convey this new semantics.

I think either of those names would be more verbose than is necessary, and would be confusing to users. The colloquial definition of what `obj must call foo` means is that `obj` needs to call `foo` along all paths, even if on some paths `obj` contains some dynamic object that doesn't really need to call `foo` (or at least that's what I mean when I say such things, and I think @msridhar and @Nargeshdb agree - though they should point out if they do not). The semantics of the `@MustCall` annotation match this meaning.

> Your arguments about being unorthodox or not enforcing or not being a type system don't make sense to me. Even if this were only a dataflow analysis, it would still need a clear definition of the dataflow lattice, in which lower lattice elements represent fewer values. Hopefully we are converging on that, so that is good news.

Fair enough. Those arguments were intended to help explain why this was difficult to explain, but they might have just made this more confusing in retrospect.

kelloggm:
@mernst I've written a README for the checker which summarizes the discussion that we've had on this PR, and which should hopefully make the intent of the type system and the type hierarchy clear. Can you take a look at that document (and the modified javadoc for `MustCall` and `MustCallAny` (formerly `MustCallTop`) and let me know if they are sufficiently clear?

must-call-checker/README.md
```diff
@@ -0,0 +1,88 @@
...
+    
+any use of `Socket` defaults to `@MustCall({"close"}) Socket`.
+
+TODO: Move the following implementation comment out of the user documentation.
+TODO: Make the `@MustCall` annotation inherited.
+
+
+### Implementation details and caveats
+
+This type system cannot be used to track a must-call obligation for a `String`, because it treats all
```
mernst, line 79: The second part of this sentence does not imply the first part.  This type system *can* track must-call obligations for strings.  Also, all String concatenations happen to have the type `@MustCall({})`.

I'm not sure the latter fact is worth calling out explicitly.  All newly-created strings have type `@MustCall({})`, and string concatenations are one example of that.

Is the problem that the AST lacks a call to `toString`?  (That is, if the toString() were explicit there would be no problem?)  The CFG contains an explicit `toString` call, so maybe you can solve your problem in the transfer function rather than by special-casing string concatenation.

mernst (changes requested):
This documentation helps a lot.  Thanks!

I have some questions about it; please see my comments.

must-call-checker/README.md
kelloggm: Maybe I should have replaced the first part of the sentence with "This type system ~~cannot~~ should not be used to track a must-call obligation for a `String`", because the special treatment of string concatenation would make doing so unsound (e.g. the result of concatenating two strings with types `@MustCall({"a"}) String` and `@MustCall({"b"}) String` would be...`@MustCall({}) String`, because all string concatenations have that type. I think that this is at least worth mentioning, since it might violate a user's expectations.

> All newly-created strings have type @MustCall({}), and string concatenations are one example of that.

It goes beyond that, which is why this is mentioned. For example, if the user for some reason used the following stub file:
```
package java.lang;
import org.checkerframework.checker.mustcall.qual.MustCall;
@MustCall("matches") class String { }
```
the result of every string concatenation would still be `@MustCall({}) String`, even though newly-created strings would have the type `@MustCall({"matches"}) String`.

> Is the problem that the AST lacks a call to toString? (That is, if the toString() were explicit there would be no problem?) The CFG contains an explicit toString call, so maybe you can solve your problem in the transfer function rather than by special-casing string concatenation.

Yes, that is the problem. I tested that explicitly when I made this fix (see the file `ToStringOnSocket.java` in the mustcall tests). Fixing it in the transfer function when encountering a `toString` call is a good idea that I didn't consider; I'll look into it.

must-call-checker/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+Note that `@MustCall({"m1", "m2"}) Object` represents more possible values than
+`@MustCall({"m1"}) Object` does, because `@MustCall({"m1"}) Object` can only
+contain an object that needs to call `m1` or one that needs to call nothing - an
+object that needs to call `m2` (or both `m1` and `m2`) is not permitted.
+
+The default type in unannotated code is `@MustCall({})`.  Given an expression of
+unannotated type, its value definitely has no obligations to call any methods.
+
+This default is unsound in the sense that a priori the checker
```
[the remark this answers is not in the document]
kelloggm: > A default cannot be unsound

I disagree with this statement. My disagreement, I think, hinges on what soundness means. 

One way to define soundness is that the type system never approves a program that is inconsistent with its annotations. I think that's the definition you are using.

Another way to define soundness is that there is no way to annotate an incorrect program with a set of consistent annotations. That's what I mean here. The problem is that for the must-call type system (and for some others we've talked about extensively, like data classification), there is no way to know a priori what the right defaults are, and taking the most conservative approach (`@MustCallAny` as the default for this type system, or `@Critical` for data classification, etc) isn't useful. Using `@MustCall({})` as the default violates this notion of soundness (but not the other notion) because we, as humans, know that some classes (like `Socket`) do actually have must-call obligations.

For many of the other type checkers we've built, this isn't a concern, because there is an objective truth that we can easily know ahead of time and build into the checker. For example, the Index Checker can know definitively what is and is not an array access, and therefore can guarantee both of these notions of soundness at the same time.

must-call-checker/README.md
```diff
@@ -0,0 +1,88 @@
...
+the documentation for this checker sometimes shortens this to "`obj` must call `foo`" or similar,
+to refer to the worst case in which `obj` is actually an object that has to call `foo` (even if
+`obj` sometimes dynamically does not actually need to call `foo`, such as when `obj` is `null`).
+
+The MustCall Checker tracks which methods might need to be called, but does
+**not** check that these methods have actually been called.  The Object
+Construction Checker's `-AcheckMustCall` command-line option checks the
+obligation expressed by an `@MustCall` annotation. This option uses a dataflow
+algorithm to compare `@MustCall` obligations to inferred `@CalledMethods` facts,
+and reports cases where required methods might not have been called.
```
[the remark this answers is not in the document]
kelloggm: > According to the documentation, no matter what @MustCall annotation appears on the type, the referent doesn't necessarily have any obligation to call any method

> Maybe the OCC is assuming a different semantics than expressed here, and the OCC assumes that @MustCall really means must call, as opposed to "might need to call".

I'll clarify this in the documentation, but the OCC isn't assuming a different semantics, but it does provide a different guarantee: the `-AcheckMustCall` command-line flag to the OCC guarantees that some set of must-call obligations were actually fulfilled. To that end, it's being conservative wrt what the MustCall checker can actually prove. If the MustCall checker can't prove that something is `@MustCall({})`, then the OCC assumes that the named methods really did need to be called, and checks them accordingly. Said another way, the MustCall checker's job is to discharge as many of these obligations as it can (by reducing them to `@MustCall({})` if possible), and the OCC checks the rest.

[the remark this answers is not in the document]
mernst: That's very helpful, thanks.

must-call-checker/README.md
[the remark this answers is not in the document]
mernst: > Another way to define soundness is that there is no way to annotate an incorrect program with a set of consistent annotations.

Can you define "incorrect program", as you mean it here?  Thanks.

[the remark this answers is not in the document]
kelloggm: I've removed this paragraph in my latest draft, as you originally suggested, based on our call earlier.

However, I think this question is independently interesting, so I'm going to try to answer it anyway. By "incorrect program" here I mean a program that does something that the checker promised wouldn't be possible if the program typechecks. The distinction I'm trying to get at between a checker like this one and e.g. the nullness or index checker is that this checker makes a guarantee **modulo a user-provided specification**, whereas those checkers make a guarantee, full stop. If the nullness checker approves your program (while being able to see the full program source) without issuing an error, the program definitely will not dereference a null pointer. By contrast, even with access to the full program text, the must-call checker (with the Object Construction Checker's `-AcheckMustCall` option) cannot notice that a required method was not called unless a user tells the checker that that method needs to be called.

must-call-checker/README.md
```diff
@@ -0,0 +1,118 @@
...
+that this approximation (the `@MustCall` type) must be fulfilled.
+
+For example, an object whose type is `java.io.OutputStream` might
+have an obligation to call the `close()` method, to release an underlying file resource. Or,
+such an `OutputStream` might not have such an obligation, if the underlying resource is, for
+example, a byte array. Either of these dynamic types can be represented by the static type
+`@MustCall({"close"}) OutputStream`, which can be read as "an OutputStream that might need
+to call close before it is deallocated". The Object Construction Checker would enforce that the
+type of such an object in its hierarchy is a subtype of `@CalledMethods({"close"})` at each
+point it is deallocated.
```
msridhar, line 19: Nit:
```suggestion
point it may become unreachable.
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -733,68 +742,91 @@ private void propagate(
   }
 
   /**
-   * Returns the String value of @AlwaysCall annotation declared on the class type of {@code
-   * element}. Returns null if the class type of {@code element} doesn't have @AlwaysCall
-   * annotation.
+   * Returns the String value of @MustCall annotation declared on the class type of {@code element}.
```
msridhar, line 745: Should we delete this?

msridhar (commented):
Just a couple minor comments, not blocking

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
kelloggm: Nice catch!

must-call-checker/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Explanation of qualifiers and type hierarchy
+
+The MustCall Checker supports two qualifiers: `@MustCall` and `@MustCallAny`. The `@MustCall`
+qualifier's arguments are the methods that the annotated type
+should call. The type `@MustCall({"a"}) Object obj` means
+that before `obj` is deallocated, `obj.a()` should be called.
+The type hierarchy is:
+
+                     @MustCallAny
```
mernst, line 29: I think this would be better named `@MustCallUnknown`.

must-call-checker/README.md
```diff
@@ -0,0 +1,118 @@
...
+they want to enforce.
+
+The programmer should rarely (if ever) write the top annotation (`@MustCallAny`), because
+the Object Construction Checker cannot verify it by matching with a corresponding `@CalledMethods`
+annotation. `@MustCallAny` theoretically represents a value with an unknowable or infinite set
+of must-call methods.
+
+### Example use
+
+If a user annotates a class declaration with a `@MustCall(...)` annotation, any use of the class's
```
mernst, line 59: Please discuss the relationship between this syntax and `@DefaultQualifierForUse`.  I think it might be the same, but I'm not sure.

must-call-checker/README.md
```diff
@@ -0,0 +1,111 @@
...
+void log(String msg, @NotOwning Object... args);
+```
+
+Now, the Object Construction Checker can soundly verify the client code without
+warnings, including both permitting the call `log()` and ensuring that `close()` is
+eventually called.
+
+### Implementation details and caveats
+
+This type system should not be used to track a must-call obligation for a `String`, because it treats all
```
mernst, line 102: I don't see why the type system should not be used to track must-call obligations for a String.

It is a fact that a newly-constructed String (whether created via `new String` or a string literal or concatenation) has no must-call obligation, but that is true of any object whose type declaration isn't annotated with `@MustCall`.  So, this does not seem like a special case to me.

Please either remove this section or explain why this is a special case and what problem it is solving, with a concrete code example.

(It occurs to me that maybe this is trying to say that string concatenation is treated as a `@NotOwning` operation.)

mernst (changes requested):
The explanation is much better, and I understand it now.  Thanks for the big improvements!

I have noted 3 places where further explanation would be helpful.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,207 @@
...
+      declaration = (ExecutableElement) TreeUtils.elementFromTree(tree);
+    } else {
+      throw new BugInCF("unexpected type of method tree: " + tree.getKind());
+    }
+    for (int i = 0; i < type.getParameterTypes().size(); i++) {
+      Element paramDecl = declaration.getParameters().get(i);
+      if (getDeclAnnotation(paramDecl, NotOwning.class) != null) {
+        AnnotatedTypeMirror paramType = type.getParameterTypes().get(i);
+        paramType.replaceAnnotation(TOP);
+        // Descend into a varargs array
```
mernst, line 77: Why does this not descend into parameterized types?

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,207 @@
...
+  @Override
+  public QualifierHierarchy createQualifierHierarchy() {
+    return new MustCallQualifierHierarchy(this.getSupportedTypeQualifiers(), this.elements);
+  }
+
+  /**
+   * The qualifier hierarchy is responsible for lub, glb, and subtyping between qualifiers without
+   * declaratively-defined subtyping relationships, like our @MustCall annotation.
+   */
+  private class MustCallQualifierHierarchy extends ElementQualifierHierarchy {
```
mernst, line 109: Please extend `MostlyNoElementQualifierHierarchy` instead.  That will enable you to simplify the implementation.

mernst (changes requested):
Here are some code suggestions.

must-call-checker/README.md
kelloggm: This is a standard syntax supported by the Checker Framework; it's only used here because it makes the example clear. Adding an explanation, in my opinion, would distract from the core of the example. From the manual (https://checkerframework.org/manual/#default-for-use):

> If a type declaration is annotated with a qualifier bound, but not a `@DefaultQualifierForUse`, then the qualifier bound is added to all unannotated uses of that type (except locals). For example, `@C class MyClass` is equivalent to:
```
@DefaultQualifierForUse(C.class)
@C class MyClass {}
```

kelloggm: Is there a specific suggestion you have on how to clarify this example that won't distract from its main point, which is to show users how to actually place an `@MustCall` annotation?

must-call-checker/README.md
kelloggm: > (It occurs to me that maybe this is trying to say that string concatenation is treated as a @NotOwning operation.)

That's one way to think of this, I think.

The larger point is that the checker is opinionated in its treatment of string concatenation, and there's no way for the user to change that. So if a user wanted to check a must-call obligation for strings, they would encounter problems related to this - that's why it is documented here.

> It is a fact that a newly-constructed String (whether created via new String or a string literal or concatenation) has no must-call obligation, but that is true of any object whose type declaration isn't annotated with @MustCall

Even if the declaration of `java.lang.String` were annotated with a non-empty `@MustCall` annotation, the result of a string concatenation would still be `@MustCall({})`. That's the reason that this section exists: string is an exception to the usual rules about newly created objects.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
kelloggm: The simple answer is that we haven't encountered a case where we needed it yet. Now that I think about it, we probably will eventually.

must-call-checker/README.md
kelloggm: Note to self, mention that this is a standard syntax and link that manual section, nothing else

PR #211, opened 2020-10-22
Add @EnsuresCalledMethodsVarArgs annotation

msridhar:
Add an `@EnsuresCalledsMethodVarArgs` annotation for methods, indicating that certain methods are called on _all_ parameters passed in the varargs position to the annotated method.  For now, this annotation is unverified and we always issue a warning when it is used.  Add support for `@EnsuresCalledMethodsVarArgs` to the transfer function of the object construction checker.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
@@ -20,7 +20,13 @@ import org.checkerframework.framework.source.SuppressWarningsPrefix;
   ObjectConstructionChecker.COUNT_FRAMEWORK_BUILD_CALLS,
   ObjectConstructionChecker.DISABLED_FRAMEWORK_SUPPORTS,
 })
-@StubFiles({"Socket.astub", "NotOwning.astub", "Stream.astub", "AlwaysCallEmpty.astub", "IOUtils.astub"})
+@StubFiles({
+  "Socket.astub",
+  "NotOwning.astub",
+  "Stream.astub",
+  "AlwaysCallEmpty.astub",
+  "IOUtils.astub"
+})
 public class ObjectConstructionChecker extends BaseTypeChecker {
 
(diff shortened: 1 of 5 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -20,7 +20,13 @@
   ObjectConstructionChecker.COUNT_FRAMEWORK_BUILD_CALLS,
   ObjectConstructionChecker.DISABLED_FRAMEWORK_SUPPORTS,
 })
-@StubFiles({"Socket.astub", "NotOwning.astub", "Stream.astub", "AlwaysCallEmpty.astub", "IOUtils.astub"})
+@StubFiles({
```
msridhar, line 23: This was an extraneous re-formatting done by Spotless

kelloggm: The `always-call-checker` branch is failing spotless right now; my branch with the MustCall Checker required the same change.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -63,6 +69,12 @@ public Properties getMessagesProperties() {
     messages.setProperty(
         "finalizer.invocation.invalid",
         "This finalizer cannot be invoked, because the following methods have not been called: %s\n");
+    messages.setProperty(
+        "ensuresvarargs.annotation.invalid",
+        "@EnsuresCalledMethodsVarArgs cannot be written on a non-varargs method");
+    messages.setProperty(
+        "ensuresvarargs.unverified",
+        "@EnsuresCalledMethodsVarArgs cannot be verified yet.  Please suppress this warning.");
```
kelloggm, line 77: "Please check that the implementation of the method actually does call the given methods on the varargs parameters by hand, and then suppress the warning."

kelloggm (approved):
Lgtm

PR #212, opened 2020-10-24
Must call polymorphism

kelloggm:
On Zookeeper, this change alone removes 48 more false positives (about 20% of the `required.method.not.called` warnings vs the always-call-before-occ branch, from 238 to 190). Hopefully this should seriously reduce the number of false positives in other benchmarks, too.


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -7,7 +7,7 @@ import org.checkerframework.common.returnsreceiver.qual.*;
 @MustCall("close")
 class DataOutputStream extends FilterOutputStream implements DataOutput {
     protected int written;
-    DataOutputStream(@Owning OutputStream arg0);
+    @PolyMustCall DataOutputStream(@Owning @PolyMustCall OutputStream arg0);
```
msridhar, line 10: I think we want to get rid of `@Owning` here?  The `DataOutputStream` does not take ownership of `arg0`; it's an "either-or" situation.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -34,7 +34,7 @@ class InputStream { }
 @MustCall("close")
 class FilterOutputStream extends OutputStream {
     protected OutputStream out;
-    FilterOutputStream(@Owning OutputStream arg0);
+    @PolyMustCall FilterOutputStream(@PolyMustCall @Owning OutputStream arg0);
```
msridhar, line 37: Get rid of `@Owning`, similar to above?

object-construction-checker/tests/mustcall/WrapperStreamPoly.java
```diff
@@ -0,0 +1,17 @@
...
+import java.io.*;
+
+class WrapperStreamPoly {
+    void test_no_close_needed(ByteArrayInputStream b) {
+        // b doesn't need to be closed, so neither does this stream.
+        DataInputStream d = new DataInputStream(b);
+    }
+
+    void test_close_needed(InputStream b) {
+        // :: error: required.method.not.called
```
msridhar, line 14: When our worklist algorithm gets smarter, this error should go away also, since formal parameters are `@NotOwning` by default

msridhar (commented):
It's like magic that this works with such little additional code!  Couple of comments.  The comments I wrote regarding `@Owning` in the stubs apply to all the stub file changes

object-construction-checker/tests/mustcall/WrapperStreamPoly.java
kelloggm: I added `@Owning` annotations to both of the parameters in this file, so this failure should stick around now.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
kelloggm: That makes sense. Do you know why these were marked as `@Owning` in the first place? That doesn't seem right to me.

kelloggm: I removed all the `@Owning` annotations on stream constructors that are also annotated with `@PolyMustCall`. Can you (@msridhar) or @Nargeshdb look over this set of changes and make sure that they make sense? After that I'll merge this PR into #207

kelloggm: It's also worth mentioning that this causes two of our tests to fail. We could make these errors expected, or we could undo the change that removes the `@Owning` annotations. I'm curious what you all think.

Here are the new failures:

1. In `WrapperStreamPoly.java`, consider this method:

   ```
    void test_close_needed(@Owning InputStream b) {
        // :: error: required.method.not.called
        DataInputStream d = new DataInputStream(b);
    }
   ```

   Removing `@Owning` from the parameter to the `DataInputStream` constructor causes a second warning to be 
   issued here, for the parameter (because ownership isn't transferred to `d`).

2. In `ACSocketTest.java`, consider this code:

   ```
    void test(String address, int port)
    {
        try
        {
            // :: error: required.method.not.called
            Socket socket = new Socket( address, 80 );

            // :: error: required.method.not.called
            PrintStream out = new PrintStream( socket.getOutputStream() );
            BufferedReader in = new BufferedReader( new InputStreamReader( socket.getInputStream() ) );
            in.close();
        }
        catch( Exception e )
        {
            e.printStackTrace();
        }
    }
   ```

   An unexpected error is issued on the line `BufferedReader in = new BufferedReader( new InputStreamReader( 
   socket.getInputStream() ) );` because ownership of the `new InputStreamReader` isn't transferred to `in`.

I don't think either of these errors are particularly desirable, and the `@Owning` annotation is only relevant when the stream actually does have a must-call obligation, so I'm not sure that removing them makes sense.

msridhar: > That makes sense. Do you know why these were marked as `@Owning` in the first place? That doesn't seem right to me.

I am not sure.  It must have just been a mistake.

msridhar: So we don't have anything in our checker yet that captures the "either-or" semantics of wrapper streams when they wrap something that needs to be closed.  I think we will add that reasoning by implementing support in the worklist algorithm for interpreting `@PolyMustCall`.  The `@Owning` annotation on the parameters of the wrapper streams does no harm for now in terms of soundness.  But given we want to add worklist-algorithm-support for `@PolyMustCall`, I am inclined to keep these errors for now, with a comment indicating we expect them to go away soon.  WDYT?

msridhar (approved):
I am ok with merging back to #207

PR #215, opened 2020-11-01
Fix a crash and add a couple optimizations

msridhar:
The crash showed up when checking Hadoop.  I think it's potentially an issue with Checker data flow.  For now, add a null check to work around it.  Also add a couple of optimizations.  The optimization to `propagate()` speeds up the worklist algorithm quite a bit in some cases.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -91,7 +91,6 @@ import org.checkerframework.javacutil.BugInCF;
 import org.checkerframework.javacutil.ElementUtils;
 import org.checkerframework.javacutil.Pair;
 import org.checkerframework.javacutil.TreeUtils;
-import org.checkerframework.javacutil.trees.DetachedVarSymbol;
 
 /**
  * The annotated type factory for the object construction checker. Primarily responsible for the
@@ -413,8 +412,6 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
           }
 
           if (lhs instanceof LocalVariableNode
-              // don't track synthetic variables
-              && !(((LocalVariableNode) lhs).getElement() instanceof DetachedVarSymbol)
               && !isTryWithResourcesVariable((LocalVariableNode) lhs)) {
 
             // Reassignment to the lhs
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -821,7 +821,7 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
 
     public BlockWithLocals(Block b, Set<LocalVarWithAssignTree> ls) {
       this.block = (BlockImpl) b;
-      this.localSetInfo = Collections.unmodifiableSet(ls);
+      this.localSetInfo = ls;
     }
 
     @Override
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -774,7 +774,9 @@ public class ObjectConstructionAnnotatedTypeFactory extends BaseAnnotatedTypeFac
 
     AnnotationMirror cmAnno;
 
-    CFValue lhsCFValue = store.getValue(assign.localVar);
+    // sometimes the store is null!  this looks like a bug in checker dataflow.
+    // TODO track down and report the root-cause bug
+    CFValue lhsCFValue = store != null ? store.getValue(assign.localVar) : null;
     if (lhsCFValue != null) { // When store contains the lhs
       cmAnno =
           lhsCFValue.getAnnotations().stream()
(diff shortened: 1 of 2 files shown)


kelloggm (approved):

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -765,13 +761,19 @@ private void propagate(
...
+      return;
+    }
     AnnotationMirror dummyCMAnno = createCalledMethods(mustCallValue.toArray(new String[0]));
 
     boolean report = true;
 
     AnnotationMirror cmAnno;
 
-    CFValue lhsCFValue = store.getValue(assign.localVar);
+    // sometimes the store is null!  this looks like a bug in checker dataflow.
```
mernst, line 774: @msridhar When I revert this hunk, the tests still pass.  It seems that the test case `MustCallNullStore.java` does not reproduce the bug.

msridhar: @mernst you may need to also comment out the emptiness check under the `// optimization` comment.  That may be masking the issue.  Sorry, I should have documented better how to reproduce.

mernst: @msridhar I cannot reproduce the problem in that way, or by going back to a previous commit and adding the `MustCallNullStore.java` file.

msridhar: Hi Mike, if I do `git checkout 39774772fbbb4c2c0382425549fbbf6f51c350ed` (which in my history is the commit before this got merged to `always-call-before-occ`), re-add the `MustCallNullStore.java` file, and then run `./gradlew :object-construction-checker:test`, I see the failure and the NPE:

```
1 unexpected diagnostic was found:
:-1: other: error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing tests/mustcall/MustCallNullStore.java
  ; The Checker Framework crashed.  Please report the crash.
  Compilation unit: tests/mustcall/MustCallNullStore.java
  Last visited tree at line 5 column 1:
  class MustCallNullStore {
  Exception: java.lang.NullPointerException; java.lang.NullPointerException
  	at org.checkerframework.checker.objectconstruction.ObjectConstructionAnnotatedTypeFactory.checkMustCall(ObjectConstructionAnnotatedTypeFactory.java:774)
  	at org.checkerframework.checker.objectconstruction.ObjectConstructionAnnotatedTypeFactory.mustCallTraverse(ObjectConstructionAnnotatedTypeFactory.java:544)
  	at org.checkerframework.checker.objectconstruction.ObjectConstructionAnnotatedTypeFactory.postAnalyze(ObjectConstructionAnnotatedTypeFactory.java:342)
  	at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericAnnotatedTypeFactory.java:1358)
```

Does that work for you?

mernst: That is exactly what I did.  Both `./gradlew test` and `./gradlew :object-construction-checker:test` work for me.  We must be doing something differently.  Could you create a new branch with your state and push it, so we can see the CI output?

msridhar: Here's the branch:

https://github.com/msridhar/object-construction-checker/tree/test-null-store-failure

Here's the failing build:

https://travis-ci.com/github/msridhar/object-construction-checker/builds/197925335

Perhaps this is an issue with using JDK 8 vs JDK 11?

mernst: I needed to run `./gradlew clean` before `./gradlew test`; otherwise `./gradlew test` passed.  This seems like a bug in the build system.  Can someone who knows Gradle better than I do look into that?

msridhar: I think this is the right line to ensure tests are re-run when a new test input is added:

https://github.com/kelloggm/object-construction-checker/blob/d94c75550da693b9b1cd02ec20e73c3984cbdcd8/object-construction-checker/build.gradle#L86

I just tested locally and when I add a new Java file under the `object-construction-checker/tests/mustcall` directory the tests get re-run even without cleaning.  So I am not sure why you need to run clean first 🤔

mernst, line 774: (Starting a new thread about this same line of code.)
I don't think this is a bug in the Checker Framework.  In your test case, the null store comes from calling getStoreAfter("return n").  There is no control flow after the return statement.  So `mustCallTraverse` could avoid calling getStoreAfter unless there are dataflow successor blocks, or it can handle the possibility that the store is null.

msridhar: It is possible this is not a bug but that we have misunderstood the spec of the `getStoreAfter` API.  However, the `return n` node in the CFG does have a successor; the relevant code calling `getStoreAfter` runs inside a conditional that checks if the successor is an exit node:

https://github.com/msridhar/object-construction-checker/blob/9d2075db86bba301179fb5585d8204babda98bb8/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java#L532-L549

This same code calls `getStoreAfter` for pretty much any `return` statement without getting back a null store.  

Also, in `MustCallNullStore.java`, if I comment out lines 15, 17, and 18 to remove the `try-finally` construct, I no longer get an NPE, which is suspicious.  If I had time to dig deeper I would try to look at how the control-flow graph looks for the `inflateDirect` method to see if there is anything suspicious there.

PR #218, opened 2020-11-02
Use CalledMethodsChecker from CF distribution rather than legacy version

kelloggm:
Some of the code (especially to handle exceptional stores) is kind of ugly, but this should work for now. I'll need to make some changes in the CF itself for that to be cleaner, but I wanted to get this working version here so that we can start using it.

Rather than preserving the old CMPredicate evaluation code, I used the default version from the CF. That uses a less-precise subtyping relationship, but we aren't using CMPredicate for MustCall anyway so I didn't think it was important. I can switch it back if a reviewer feels strongly.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -51,89 +54,41 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
   @Override
   public TransferResult<CFValue, CFStore> visitMethodInvocation(
       final MethodInvocationNode node, final TransferInput<CFValue, CFStore> input) {
+
+    exceptionalStores = makeExceptionalStores(node, input);
     TransferResult<CFValue, CFStore> result = super.visitMethodInvocation(node, input);
+    Map<TypeMirror, CFStore> exceptionalStoresTmp = exceptionalStores;
```
msridhar, line 60: What is going on with this temp variable?

msridhar (commented):
Overall looks ok.  Was the idea to keep all the existing test cases even if they are redundant with the called methods checker tests, just to ensure no breakage?  Also one question:

msridhar:
> Rather than preserving the old CMPredicate evaluation code, I used the default version from the CF. That uses a less-precise subtyping relationship, but we aren't using CMPredicate for MustCall anyway so I didn't think it was important. I can switch it back if a reviewer feels strongly.

This is fine.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
kelloggm: `super.visitMethodInvocation` eventually leads to calls to `accumulate`, which is overridden to side-effect `exceptionalStores`. Because `makeExceptionalStores` needs a `TransferInput`, it has to be set here, not in `accumulate`, where it is used - `accumulate` only has access to the `TransferResult`.

I realized looking back at this that I could rewrite this to avoid the temporary variable. It should be a bit clearer now; I also added some documentation to the field that explains why it has to be `@Nullable`.

kelloggm:
> Was the idea to keep all the existing test cases even if they are redundant with the called methods checker tests, just to ensure no breakage?

Yes, that was the idea.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -175,6 +138,40 @@ private void handleEnsuresCalledMethodVarArgs(
...
+              // modifiable list.
+              oldFlowValues.addAll(valuesAsList);
+              valuesAsList = oldFlowValues;
+            }
+          }
+        }
+      }
+    }
+    AnnotationMirror newAnno = typeFactory.createAccumulatorAnnotation(valuesAsList);
+    exceptionalStores.values().stream().forEach(s -> s.insertValue(target, newAnno));
```
msridhar, line 172: Is the eventual fix to have the called methods checker itself propagate to exceptional stores in this manner?  Or is there still some discussion to be had as to whether that is the right thing to do?

msridhar (commented):
LGTM.  Your call on whether to merge before or after #207

msridhar (approved):

mernst (approved):
It's great to see so much code deleted.

PR #219, opened 2020-11-02
Check fields

Nargeshdb:
This PR implements ownership transfer for owning fields. For any assignment to an owning field:
- If the field is final, then we can enforce this property with two checks on field f:
enclosing element of the field must (like class C) have an annotation @MustCall(“m2”) for some m2
C.m2() must be annotated @EnsuresCalledMethods(value = “this.f”, method=”m1”)
- If the field is non-final, then we report a warning because can't prove if the field gets overwritten or not.

We don't have ownership transfer to a field without owning annotation.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -554,6 +576,66 @@ private void mustCallTraverse(ControlFlowGraph cfg) {
...
+    Node lhs = assignmentNode.getTarget();
+    Node rhs = assignmentNode.getExpression();
+    Element fieldElement = TreeUtils.elementFromTree(lhs.getTree());
+    Element enclosingElement = fieldElement.getEnclosingElement();
+    List<String> rhsMCAnno = getMustCallValue(TreeUtils.elementFromTree(rhs.getTree()));
+    AnnotationMirror rhsCalledMethodAnno = createCalledMethods(rhsMCAnno.toArray(new String[0]));
+    List<String> enclosingElAnno = getMustCallValue(enclosingElement);
+    boolean report = true;
+    if (enclosingElAnno != null
+        && (getMustCallValue(fieldElement).isEmpty()
```
msridhar, line 589: Why do we check `getMustCallValue(fieldElement).isEmpty()`?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -554,6 +576,66 @@ private void mustCallTraverse(ControlFlowGraph cfg) {
...
+      for (Element element : classElements) {
+        if (element.getKind().equals(ElementKind.METHOD)) {
+          // TODO cause false positives if there is more than one value in the enclosingElAnno
+          if (element.getSimpleName().toString().equals(enclosingElAnno.get(0))) {
+            AnnotationMirror annotationMirror =
+                getDeclAnnotation(element, EnsuresCalledMethods.class);
+            if (annotationMirror == null) {
+              break;
+            }
+            Attribute values = ((Attribute.Compound) annotationMirror).getValue().values.get(0).snd;
```
msridhar, line 601: Please place some of this extraction logic into separate methods with appropriate names.  This is pretty hard to read.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -554,6 +576,66 @@ private void mustCallTraverse(ControlFlowGraph cfg) {
...
+              Attribute methods =
+                  ((Attribute.Compound) annotationMirror).getValue().values.get(1).snd;
+              Set<String> setOfMethod =
+                  ((List<Attribute>) methods.getValue())
+                      .stream()
+                          .map(method -> method.getValue().toString())
+                          .collect(Collectors.toSet());
+              AnnotationMirror ensuresCalledMethodAnno =
+                  createCalledMethods(setOfMethod.toArray(new String[0]));
+              if (getQualifierHierarchy().isSubtype(ensuresCalledMethodAnno, rhsCalledMethodAnno)) {
```
msridhar, line 612: Why is this doing a check comparing against `rhsCalledMethodAnno`?  I think we should only be checking that the `@EnsuresCalledMethod` annotation has all methods in the `@MustCall` annotation for the _declared type_ of the field.  It should be independent of any assignment into the field.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -554,6 +576,66 @@ private void mustCallTraverse(ControlFlowGraph cfg) {
...
+      }
+    }
+
+    if (report) {
+      if (fieldElement.getModifiers().contains(Modifier.FINAL)) {
+        checker.reportError(
+            fieldElement,
+            "required.method.not.called",
+            TreeUtils.typeOf(rhs.getTree()).toString(),
+            " required methods are not called on this field");
```
msridhar, line 628: We should give more specific error messages for each case

msridhar (changes requested):
Overall the checks look good.  But we need to restructure this code a bit and move some checks to different places

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -411,6 +415,24 @@ private void mustCallTraverse(ControlFlowGraph cfg) {
...
+          Element lhsElement = TreeUtils.elementFromTree(lhs.getTree());
+
+          if (lhsElement.getKind().equals(ElementKind.FIELD) && hasMustCall(lhs.getTree())) {
+            List<String> rhsMustCallVal = getMustCallValue(rhs.getTree());
+
+            if (getDeclAnnotation(lhsElement, Owning.class) != null) {
+              // TODO nullness checking
+              checkAssignmentToOwningField((AssignmentNode) node);
+              if (rhs instanceof LocalVariableNode) {
+                if (hasMustCall(lhs.getTree()) && isVarInDefs(newDefs, (LocalVariableNode) rhs)) {
```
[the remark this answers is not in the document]
Nargeshdb: We don't need to check it! I'll remove it.

PR #222, opened 2020-11-03
Add @InheritableMustCall annotation

kelloggm:


msridhar:
Can we now remove some unnecessary `@MustCall` annotations in some stub files?

msridhar (changes requested):
Looks good to me, but I think we should also delete any `@MustCall` annotations in stubs that are now unnecessary and make sure the tests still pass

kelloggm:
> Looks good to me, but I think we should also delete any @MustCall annotations in stubs that are now unnecessary and make sure the tests still pass

I think I've removed all the redundant ones, and the tests look like they're still passing.

msridhar (approved):

PR #223, opened 2020-11-04
Extract and refactor code that ensures @MustCall methods are invoked

msridhar:
* Extract the worklist algorithm for ensuring `@MustCall` methods are invoked to a new class `MustCallInvokedChecker`
* Break the code down into smaller methods
* Unify handling of successors.  Before we had separate methods for handling exceptional vs. normal successors, which led to some ugliness and bugs.  We now have a `handleSuccessors()` method for uniformly handling successors.  This change allowed us to remove a hack that @kelloggm was concerned about where we would eagerly remove variables from the tracked set without consulting the `@CalledMethods` types.  There is one additional expected error in our unit tests (see `ACRegularExitPointTest`), as this change removes a bit of path sensitivity (which I don't think is relevant in practice).

Further cleanup is likely possible, and further documentation would be useful.  But I wanted to put up this change now since it's a good baseline for further changes, and it is likely to cause more and more merge conflicts if we try to make it too complete in one go.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -1,0 +1,551 @@
+package org.checkerframework.checker.objectconstruction;
+
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Type;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Name;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+import org.checkerframework.checker.calledmethods.qual.CalledMethods;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.objectconstruction.qual.Owning;
+import org.checkerframework.common.basetype.BaseTypeChecker;
+import org.checkerframework.dataflow.cfg.ControlFlowGraph;
+import org.checkerframework.dataflow.cfg.UnderlyingAST;
+import org.checkerframework.dataflow.cfg.block.Block;
+import org.checkerframework.dataflow.cfg.block.BlockImpl;
+import org.checkerframework.dataflow.cfg.block.ConditionalBlock;
+import org.checkerframework.dataflow.cfg.block.ExceptionBlock;
+import org.checkerframework.dataflow.cfg.block.ExceptionBlockImpl;
+import org.checkerframework.dataflow.cfg.block.RegularBlockImpl;
+import org.checkerframework.dataflow.cfg.block.SingleSuccessorBlock;
+import org.checkerframework.dataflow.cfg.block.SpecialBlock;
+import org.checkerframework.dataflow.cfg.block.SpecialBlockImpl;
+import org.checkerframework.dataflow.cfg.node.AssignmentNode;
+import org.checkerframework.dataflow.cfg.node.LocalVariableNode;
+import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
+import org.checkerframework.dataflow.cfg.node.Node;
+import org.checkerframework.dataflow.cfg.node.ObjectCreationNode;
+import org.checkerframework.dataflow.cfg.node.ReturnNode;
+import org.checkerframework.dataflow.cfg.node.TypeCastNode;
+import org.checkerframework.dataflow.expression.LocalVariable;
+import org.checkerframework.framework.flow.CFAnalysis;
+import org.checkerframework.framework.flow.CFStore;
+import org.checkerframework.framework.flow.CFValue;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.BugInCF;
+import org.checkerframework.javacutil.TreeUtils;
+
+/**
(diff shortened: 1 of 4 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -0,0 +1,515 @@
...
+import org.checkerframework.framework.flow.CFStore;
+import org.checkerframework.framework.flow.CFValue;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.TreeUtils;
+
+/**
+ * Checks that all methods in {@link org.checkerframework.checker.mustcall.qual.MustCall} object
+ * types are invoked before the corresponding objects become unreachable
+ */
+class MustCallInvokedChecker {
```
kelloggm, line 50: Nit: I hate that not putting a modifier means "package-private" in Java, because it's easy to miss. So as a matter of style I usually add `/* package-private */ where the modifier would otherwise be.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -0,0 +1,515 @@
...
+  }
+
+  private void handleInvocation(Set<LocalVarWithTree> newDefs, Node node) {
+    List<Node> arguments;
+    ExecutableElement executableElement;
+    if (node instanceof MethodInvocationNode) {
+      MethodInvocationNode invocationNode = (MethodInvocationNode) node;
+      arguments = invocationNode.getArguments();
+      executableElement = TreeUtils.elementFromUse(invocationNode.getTree());
+    } else {
```
kelloggm, line 123: I'd add a test here that this is actually an `ObjectConstructionNode`, and throw a `BugInCF` in the actual `else` block, just to be defensive.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -0,0 +1,515 @@
...
+      executableElement = TreeUtils.elementFromUse(((ObjectCreationNode) node).getTree());
+    }
+
+    List<? extends VariableElement> formals = executableElement.getParameters();
+    if (arguments.size() != formals.size()) {
+      // this could happen, e.g., with varargs, or with strange cases like generated Enum
+      // constructors
+      // for now, just skip this case
+      // TODO allow for ownership transfer here if needed in future
+      return;
```
kelloggm, line 134: We don't have to address this now, but it seems to indicate that there's currently a soundness hole in our ownership transfer algorithm if a varargs parameter is marked as `@Owning`. We should probably add validation that there isn't such an `@Owning` annotation and issue a warning if there is in the visitor.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -0,0 +1,515 @@
...
+
+  /**
+   * Is {@code exceptionClassName} an exception type we are ignoring, to avoid excessive false
+   * positives? For now we ignore {@code java.lang.Throwable} and {@code NullPointerException}
+   */
+  private static boolean isIgnoredExceptionType(Name exceptionClassName) {
+    boolean isThrowableOrNPE =
+        exceptionClassName.contentEquals(Throwable.class.getSimpleName())
+            || exceptionClassName.contentEquals(NullPointerException.class.getSimpleName());
+    return isThrowableOrNPE;
```
kelloggm, line 429: Will this also ignore user-written exceptions with these names? If so, that's a soundness issue and we should check the fully-qualified name instead..

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -0,0 +1,515 @@
...
+            outOfScopeReason);
+      }
+    }
+  }
+
+  /**
+   * Is {@code exceptionClassName} an exception type we are ignoring, to avoid excessive false
+   * positives? For now we ignore {@code java.lang.Throwable} and {@code NullPointerException}
+   */
+  private static boolean isIgnoredExceptionType(Name exceptionClassName) {
```
kelloggm, line 425: Can we add a Signature Checker annotation to `exceptionClassName` to indicate the kind of name we're expecting here? I find those annotations very useful just as documentation, even if we aren't running the signature checker (yet?).

kelloggm (changes requested):

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: c7bb347

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: e48f172

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: b4a42bb

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: caba8d5

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: I don't think this is a soundness hole.  We will just fail to transfer ownership to the varargs formal parameter, which means we will get an error somewhere else if the `@MustCall` obligation is not satisfied.  I agree we could add a warning if someone writes `@Owning` on a varargs parameter to make this more evident to the user.  I will add to my TODO list

kelloggm (approved):

PR #226, opened 2020-11-06
Align borrowing logic between MustCall checker and OCC worklist algorithm for method parameters

kelloggm:
I wasn't able to get `@Owning` receivers to work correctly, though, so this code isn't sound wrt them. Two questions:
* does it make sense to have an `@Owning` receiver, or is it okay to just not check receivers at all?
* does anyone know how to get an element representing the receiver parameter of a method from the `ExecutableElement`, if it exists? As far as I can tell there's just not one, but that seems...unlikely. @mernst?

I also removed some now-extraneous `@MustClose` annotations that are redundant with the `@InheritableMustCall("close")` annotation on `Closeable`.


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -25,6 +27,29 @@ public MustCallVisitor(BaseTypeChecker checker) {
     super(checker);
   }
 
+  @Override
+  protected boolean skipReceiverSubtypeCheck(
+      MethodInvocationTree node,
+      AnnotatedTypeMirror methodDefinitionReceiver,
+      AnnotatedTypeMirror methodCallReceiver) {
+    // TODO: Check explicit receiver parameters annotated with @Owning. ExecutableElement
```
msridhar, line 35: Does it work to get the annotations on the receiver type, rather than from the element?

msridhar:
Plan to review once CI passes.  If you need a review sooner let me know

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
kelloggm: No. I did also try that :)

msridhar: Ok.  I think we're fine to just never allow ownership transfer for receivers for now

kelloggm:
@msridhar the CI failures are caused by the bug I just fixed in https://github.com/typetools/checker-framework/pull/3867; they won't go away until it's merged and released (so, next month). I'm going to suppress the warning for now, and we can remove the suppression when we update to 3.8.0.

msridhar (approved):

PR #232, opened 2020-11-13
Revamp logic for checking that calls get pseudo-assigned to an owning location

msridhar:
The checking that `@MustCall` obligations are satisfied must ensure that a method or constructor call with non-empty `@MustCall` type gets assigned to some variable / field that takes ownership of the value.  This PR revamps the logic for this check, moving it out of the `ObjectConstructionVisitor` and instead leveraging the `AssignmentContext`s available for CFG nodes.  This change handles more cases in a uniform manner and also sets us up for future enhancements (like leveraging `@PolyMustCall` / `@MustCallChoice` annotations).

This is a WIP since it relies on https://github.com/typetools/checker-framework/pull/3891 for proper handling of method calls appearing as a parameter to a constructor call.  The code is currently written against the latest Checker Framework main branch, which required a couple of other changes.  If we use the 3.7.1 CF release, we will get false positives that we will have to ignore until the next release (or we'll have to add some extra temporary hacks / workarounds).  If we code against the CF main branch (assuming https://github.com/typetools/checker-framework/pull/3891 lands), it will get hard to get CI to pass.  I'm not sure of the best way forward.


kelloggm:
@msridhar I'm fine with switching to CF head and just letting CI fail for a few weeks. We'll just all have to be diligent that local tests are passing until 3.8.0 is released.

msridhar:
Ok, thanks, @kelloggm.  In that case let's see if https://github.com/typetools/checker-framework/pull/3891 lands soon, and if it does we can switch over to CF head

msridhar:
I think this can be reviewed now, actually.  To test locally, build the CF branch https://github.com/msridhar/checker-framework/tree/assignment-context-for-object-creation and install it to Maven local with `./gradlew deployArtifactsToLocalRepo`, and then run the tests here.

msridhar:
Update: https://github.com/typetools/checker-framework/pull/3891 got merged to CF master, so there's no need to use a different CF branch to test anymore

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -111,14 +116,143 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
+    doOwnershipTransferToParameters(newDefs, node);
+    // If the method call is nested in a type cast, we won't have a proper AssignmentContext for
+    // checking.  So we defer the check to the corresponding TypeCastNode
+    if (!nestedInTypeCast(node) && !shouldSkipInvokePseudoAssignCheck(node.getTree())) {
+      checkPseudoAssignToOwning(node);
+    }
+  }
+
+  /**
+   * Give a node representing a method or constructor call, checks that if the call has a non-empty
```
kelloggm, line 147: give -> given

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -111,14 +116,143 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
+  }
+
+  /**
+   * Checks for cases where we do not need to ensure that a method invocation gets pseudo-assigned
+   * to a variable / field that takes ownership. We can skip the check when the invoked method's
+   * return type is {@link org.checkerframework.common.returnsreceiver.qual.This}, the invocation is
+   * a super constructor call, or the method's return type is annotated {@link NotOwning}
+   */
+  private boolean shouldSkipInvokePseudoAssignCheck(Tree callTree) {
+    if (callTree instanceof MethodInvocationTree) {
```
kelloggm, line 219: I think it's preferred to compare `TreeKind`s rather than do `instanceof` on `Tree`s. So this check would become `callTree.getKind == TreeKind.METHOD_INVOCATION`.

kelloggm: I don't remember why, though.

kelloggm (approved):
Overall these changes look great. I was able to get the tests to pass locally following your instructions. I left a few minor comments, but there's no need for me to review this again.

Wrt merging, do we want to merge this into `always-call-checker` and allow CI to fail until 3.8.0 is released in December? That's probably the right move, even if it is a bit annoying. We'll just have to be careful about testing changes locally...

PR #237, opened 2020-11-17
Treat non-owning method parameters as @MustCall({}) while typechecking the body of the method

kelloggm:
~~The implementation is kind of hacky and relies on there being no other identifiers in a method body with the same name as its parameters. I *think* the scoping rules make that true, but I'm not sure.~~

I tried a couple of ~~less hacky~~ _other_ ways to implement this, but they all failed. Here are the ones that were most promising:
* override `TypeAnnotator#visitExecutable` and `GenericAnnotatedTypeFactory#addComputedTypeAnnotations(Element, ...)`: also changes the types when the method is called, which causes lots of false positives.
* override `visitVariable` in a `TreeAnnotator`, and change the type of the variable (both in the caches and the parameter) if the variable being declared is a non-owning method parameter. Doesn't seem to do anything except cause polymorphic types to be resolved incorrectly.
* override `visitMethod` in a `TreeAnnotator`, since we really only want this change to take effect while we typecheck the method body. Doesn't do anything.

If a reviewer has a better way to do this, I'm very open to trying it!

Test changes are all because the relevant parameters were implicitly `@Owning` before, because the code is artificial and intended only to test the Must Call Checker.

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
@@ -1,8 +1,11 @@
 package org.checkerframework.checker.mustcall;
 
+import static javax.lang.model.element.ElementKind.CONSTRUCTOR;
+import static javax.lang.model.element.ElementKind.METHOD;
 import static org.checkerframework.common.value.ValueCheckerUtils.getValueOfAnnotationWithStringArgument;
 
 import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.IdentifierTree;
 import com.sun.source.tree.MemberReferenceTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.NewClassTree;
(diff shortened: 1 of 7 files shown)


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -233,4 +252,45 @@ protected boolean isSubtypeWithElements(
...
+      super(mustCallAnnotatedTypeFactory);
+    }
+
+    // When they appear in the body of a method or constructor, treat non-owning parameters
+    // as bottom regardless of their declared type.
+    @Override
+    public Void visitIdentifier(IdentifierTree node, AnnotatedTypeMirror type) {
+      Element elt = TreeUtils.elementFromTree(node);
+      /// This code derived from ElementUtils#enclosingClass
+      Element enclosing = elt;
```
msridhar, line 267: Does it work to check if `elt.getKind() == ElementKind.PARAMETER` to see if the identifier is a parameter?

msridhar (commented):
Yeah, this is actually a bit tricky.  The issue is that we want `@Owning` formal parameters to be treated as `@MustCallUnknown` at call sites but `@MustCall({})` within the method body.  I don't know if there's a cleaner way than checking at uses of the formal.  But one suggestion / question below:

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
kelloggm: This is a good idea, and it does help with my concerns about scoping. I added a check at the beginning of the method that returns early if the identifier isn't a parameter of the method.

Even with that change we can't remove anything from the rest of the method, though. We need to find the formal parameter that has the same name as the element, because that's the only place we can check for an `@Owning` annotation.

kelloggm:
@mernst, it would be great if you can review this (especially the override of `visitIdentifier`). The build is failing because #232, which was merged earlier today, relies on CF 3.8.0. Until 3.8.0 is released, all our CI builds will fail. I've tested this build on my machine and it passes with CF master.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
msridhar: So calling `getDeclAnnotation(elt, Owning.class)` doesn't work?  That's surprising to me.  I would think that `elt` _is_ the element for the formal parameter.

kelloggm: You're right, it is. I must not have had enough caffeine yet today. The code is now much, much simpler - thanks!

msridhar (approved):

PR #242, opened 2020-11-24
MustCallChoice

Nargeshdb:
This PR handles @MustCallChoice annotation to reason about either-or scenarios described [here](https://docs.google.com/document/d/18m_hYV_cU738pAPBilVoFy1ecRSu5mKeVIFRCCUsyQk/edit?usp=sharing).

This is a work in progress.


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -64,9 +65,20 @@ public MustCallAnnotatedTypeFactory(final BaseTypeChecker checker) {
     BOTTOM = createMustCall();
     POLY = AnnotationBuilder.fromClass(elements, PolyMustCall.class);
     addAliasedAnnotation(InheritableMustCall.class, MustCall.class, true);
+    addAliasedAnnotation(MustCallChoice.class, POLY);
```
msridhar, line 68: What was the final strategy we went with for the `@MustCallChoice` annotation?  As an alias, or as a separate declaration annotation?

must-call-checker/tests/mustcall/MustCallChoiceImpl.java
```diff
@@ -0,0 +1,22 @@
+// A simple test that the extra obligations that MustCallChoice imposes are
+// respected.
+
+// @skip-test until the checks are implemented
+
+import org.checkerframework.checker.mustcall.qual.*;
+import org.checkerframework.checker.objectconstruction.qual.*;
+import java.io.*;
+
+public class MustCallChoiceImpl implements Closeable {
```
msridhar, line 10: It would be good to write a version of this test also for the OCC.  For that I think we'd need to add a `@MustCall` annotation on the type and `@EnsuresCalledMethods` on `close()`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -93,9 +97,9 @@
   void checkMustCallInvoked(ControlFlowGraph cfg) {
     // add any owning parameters to initial set of variables to track
     BlockWithLocals firstBlockLocals =
-        new BlockWithLocals(cfg.getEntryBlock(), computeOwningParameters(cfg));
+        new BlockWithLocals(cfg.getEntryBlock(), (computeOwningParameters(cfg)));
```
msridhar, line 100: These extra parentheses are redundant, no?  If so, let's remove this change

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -148,25 +152,27 @@ private void handleInvocation(Set<LocalVarWithTree> newDefs, Node node) {
...
       if (assignmentContext instanceof AssignmentLhsContext) {
         // lhs should be a local variable
-        assignedToOwning = isOwningAssignmentLhs(elementForType);
+        assignedToOwningOrMustCallChoice = isOwningAssignmentLhs(elementForType);
       } else if (assignmentContext instanceof MethodParameterContext) {
         // must be an @Owning parameter
-        assignedToOwning = typeFactory.getDeclAnnotation(elementForType, Owning.class) != null;
+        assignedToOwningOrMustCallChoice =
+            typeFactory.getDeclAnnotation(elementForType, Owning.class) != null
+                || typeFactory.hasMustCallChoice(elementForType);
```
msridhar, line 172: So the idea here is if the formal parameter is `@MustCallChoice`, we "pretend" the call was assigned to an owning location, since we can satisfy the `@MustCall` obligation on the returned value from the call; is that right?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -178,8 +184,14 @@ private void checkPseudoAssignToOwning(Node node) {
       } else {
         throw new BugInCF("unexpected AssignmentContext type " + assignmentContext.getClass());
       }
+    } else {
+      Node mustCallChoiceParam = getMustCallChoiceParam(node, defs);
```
msridhar, line 188: Can you add a comment to explain what is going on in this case?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +339,135 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
-        newDefs.remove(latestAssignmentPair);
+        if (setContainsLatestAssignmentPair.size() == 1) {
+          checkMustCall(
+              newDefs,
+              setContainsLatestAssignmentPair,
+              typeFactory.getStoreBefore(node),
+              "variable overwritten by assignment " + node.getTree());
+          newDefs.remove(setContainsLatestAssignmentPair);
+        } else {
+          setContainsLatestAssignmentPair.remove(latestAssignmentPair);
```
msridhar, line 364: This mutation of `setContainsLatestAssignmentPair` is dangerous!  It's going to change the set everywhere, including any place it is stored in the `visited` set.  I think we need to change `BlockWithLocals` to only store `ImmutableSet`s, and then create a fresh set here with the value removed

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +339,135 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
       // If the rhs is an ObjectCreationNode, or a MethodInvocationNode, then it adds
       // the AssignmentNode to the newDefs.
       if ((rhs instanceof ObjectCreationNode)
           || (rhs instanceof MethodInvocationNode
               && !hasNotOwningReturnType((MethodInvocationNode) rhs))) {
-        newDefs.add(
-            new LocalVarWithTree(new LocalVariable((LocalVariableNode) lhs), node.getTree()));
+        if (typeFactory.hasMustCallChoice(rhs.getTree())) {
+          Node n = getMustCallChoiceParam(rhs, newDefs);
+          if (n != null && isVarInDefs(newDefs, (LocalVariableNode) n)) {
```
msridhar, line 375: How do we know that `n` will be a `LocalVariableNode`?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +339,135 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+            new LocalVarWithTree(new LocalVariable((LocalVariableNode) lhs), node.getTree());
+        Set<LocalVarWithTree> newSet = new LinkedHashSet<>();
+        newSet.add(lhsLocalVarWithTreeNew);
+        newDefs.add(newSet);
+        newDefs.remove(setContainsRhs);
+      }
+    }
+  }
+
+  private Node getMustCallChoiceParam(Node node, Set<Set<LocalVarWithTree>> defs) {
```
msridhar, line 407: Need to document this method, and if it always returns `@Nullable LocalVariableNode` then change the declared type

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -605,9 +731,9 @@ static String formatMissingMustCallMethods(List<String> mustCallVal) {
    */
   private static class BlockWithLocals {
     public final Block block;
-    public final Set<LocalVarWithTree> localSetInfo;
+    public final Set<Set<LocalVarWithTree>> localSetInfo;
```
msridhar, line 734: This field should have a type like `ImmutableSet<ImmutableSet<LocalVarWithTree>>`.  That will force us to be a lot more careful with mutation elsewhere.

msridhar (commented):
Some initial comments

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
kelloggm: It is an alias. In the end that was the only strategy that I could get working reliably.

object-construction-checker/tests/socket/InputOutputStreams.java
```diff
@@ -0,0 +1,100 @@
...
+            OutputStream os = sock.getOutputStream();
+        } catch (IOException e) {
+
+        } finally {
+            sock.close();
+        }
+    }
+
+    // :: error: required.method.not.called
+    void test_close_is(@Owning Socket sock) throws IOException {
```
kelloggm, line 21: The reason that this test (and several of the following) should fail is subtle, because on first glance they look correct. Can we add a comment about how `getInputStream` might throw an exception, making `is` null in the finally block - thus causing the socket not to be closed? I think we only need that note once - it also explains why e.g. `test_close_os` below should result in a warning. I'm asking because I screwed that up when I originally wrote these tests :)

object-construction-checker/tests/socket/InputOutputStreams.java
```diff
@@ -0,0 +1,100 @@
...
+            try {
+            os = sock.getOutputStream();
+            } catch (IOException ee) {}
+        } finally {
+            os.close();
+        }
+    }
+
+    // :: error: required.method.not.called
+    void test_close_os3(@Owning Socket sock) throws IOException {
```
kelloggm, line 57: Can we add this version of this test:

```
void test_close_os4(@Owning Socket sock) throws IOException {
        OutputStream os = null;
        try {
            InputStream is = sock.getInputStream();
        } catch (IOException e) { }
        try {
            os = sock.getOutputStream();
        } finally {
            if (os != null) {
                os.close();
            } else {
                sock.close();
            }
        }
    }
```

I think this version should pass.

kelloggm (changes requested):
Some minor comments on the tests, mostly. I'll review the code itself once Manu's review is complete - please ping me then.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: exactly!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: In this case, we are handling method invocation nodes that are not assigned to anything but the receiver of the method is in the defs. For example, we won't report a warning for "sockChannel.socket();" if sockChannel is a local variable in the defs.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: Because getMustCallChoiceParam() returns LocalVariableNode if any exists, or returns null.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -514,41 +636,57 @@ private static boolean isTryWithResourcesVariable(LocalVariableNode lhs) {
...
-      cmAnno =
-          typeFactory
-              .getAnnotatedType(localVarWithTree.localVar.getElement())
-              .getAnnotationInHierarchy(typeFactory.top);
-    }
-
-    if (!calledMethodsSatisfyMustCall(mustCallValue, cmAnno)) {
-      if (!reportedMustCallErrors.contains(localVarWithTree)) {
-        reportedMustCallErrors.add(localVarWithTree);
+    Set<LocalVarWithTree> defsCopy = new LinkedHashSet<>(localVarWithTreeSet);
```
msridhar, line 652: Why do we need the `defsCopy` and do `iter.remove()` at the end of the loop?  That looks unnecessary to me

msridhar (changes requested):
A few things to fix

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: I think it's better to keep this check here because we need to know if the method call or object creation node is passed to a @MustCallChoice parameter or not! So, with AssignmentContext we can do this check! what do you think about changing the name of checkPseudoAssignToOwning method instead?!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,46 +344,144 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
           || (rhs instanceof MethodInvocationNode
               && !hasNotOwningReturnType((MethodInvocationNode) rhs))) {
-        newDefs.add(
-            new LocalVarWithTree(new LocalVariable((LocalVariableNode) lhs), node.getTree()));
+        if (typeFactory.hasMustCallChoice(rhs.getTree())) {
+          LocalVariableNode n = getMustCallChoiceParam(rhs);
+          if (n != null && isVarInDefs(newDefs, n)) {
+            Set<LocalVarWithTree> immutableSet = getSetContainsAssignmentTreeOfVar(newDefs, n);
+            Set<LocalVarWithTree> newImmutableSet = new LinkedHashSet<>(immutableSet);
+            FluentIterable.from(newImmutableSet)
```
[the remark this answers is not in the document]
Nargeshdb: This part of the code didn't work as I expected! I updated it! Thanks for catching this bug.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: Sorry I wrote this comment in the wrong place!  Nevermind 🙂

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -178,8 +188,15 @@ private void checkPseudoAssignToOwning(Node node) {
       } else {
         throw new BugInCF("unexpected AssignmentContext type " + assignmentContext.getClass());
       }
+    } else {
+      // In this case, we are handling method invocation nodes that are not assigned to a local
+      // variable node but the receiver of the method is in the defs
+      LocalVariableNode mustCallChoiceParam = getMustCallChoiceParam(node);
```
msridhar, line 194: _This_ is the check that should be moved to `shouldSkipInvokePseudoAssignCheck` (I think)

object-construction-checker/tests/mustcall/MustCallChoiceExamples.java
```diff
@@ -0,0 +1,54 @@
...
+            socket = new Socket( address, 80 );
+            DataInputStream d = new DataInputStream(socket.getInputStream());
+        } catch (IOException e){
+
+        } finally {
+            closeSocket(socket);
+        }
+    }
+
+    // :: error: required.method.not.called
```
msridhar, line 23: We should not get an error here!  Why can't our tool prove this one safe?  (Note that `new DataInputStream(b)` cannot throw an `IOException`)

object-construction-checker/tests/mustcall/MustCallChoiceImpl.java
```diff
@@ -0,0 +1,19 @@
+//import org.checkerframework.checker.mustcall.qual.*;
```
msridhar, line 1: Why is this all commented out?

object-construction-checker/tests/socket/ACSocketTest.java
```diff
@@ -479,7 +477,13 @@ private SSLSocket configureSSLSocket(@Owning SSLSocket socket, boolean isClientS
...
 
     private void updateSocketAddresses( SelectionKey sockKey ) {
-        // error: required.method.not.called
+        Socket socket = ((SocketChannel) sockKey.channel()).socket();
+        SocketAddress localSocketAddress = socket.getLocalSocketAddress();
+        SocketAddress remoteSocketAddress = socket.getRemoteSocketAddress();
+    }
+
+    private void updateSocketAddressesWithOwning(@Owning SelectionKey sockKey ) {
+        // TODO we should report a warning here
```
msridhar, line 486: Why don't we report a warning now?  Please explain

msridhar (commented):
Couple more comments

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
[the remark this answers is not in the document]
msridhar: Interestingly I think this checker itself could have caught this bug 🙂 Roughly, various `Stream` operations should actually return a `@MustCall("collect") Stream` (or something like that)

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: You're right, I'll remove it.

object-construction-checker/tests/mustcall/MustCallChoiceExamples.java
Nargeshdb: you're right. I fixed it.

object-construction-checker/tests/socket/InputOutputStreams.java
Nargeshdb: It's a great example! To pass this case, we should handle the boolean condition problem.

msridhar: To further clarify, in the case where `getOutputStream()` throws an `IOException`, `os` does _not_ get added to the set of "choice" variables for the socket resource.  Further, the `os` variable remains definitely `null`.  When our worklist analysis gets to the `finally` block along the corresponding CFG edge, it does not know that `os` is definitely null, and it is only tracking `sock` as a name for the resource.  So, it analyzes a path through the "then" branch of the conditional, where `sock.close()` is not invoked, and reports an error.  Addressing this issue requires more general tracking of additional boolean conditions (in this case, regarding the nullness of `os`), which we should leave outside the scope of this pull request.

kelloggm: Thanks for the explanation. If possible, I'd like to keep the test case with a note about why it fails (probably exactly what you wrote in the comment above) - I suspect it will be useful later.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +362,153 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+   * object creation node: 1) If there is a formal parameter with {@link @MustCallChoice}
+   * annotation, it passes the parameter to a recursive call to handle nested calls 2) If there is
+   * no formal parameter with {@link @MustCallChoice} annotation, then it checks the receiver
+   * parameter. If the receiver parameter has @MustCallChoice annotation, then it passes the
+   * receiver parameter to a recursive call to handle sequence of method calls
+   *
+   * @param node
+   * @return {@code node} iff {@code node} represents a local variable, otherwise null
+   */
+  private @Nullable LocalVariableNode getMustCallChoiceParam(Node node) {
```
msridhar, line 455: Let's rename the method to `getLocalPassedAsMustCallChoiceParam()`.  The documentation needs to be improved.  Basically, this method tries to find a local variable passed as a `@MustCallChoice` parameter.  In the base case, if `node` is a local variable, it just gets returned.  Otherwise, if `node` is a call (or a call wrapped in a cast), the code finds the parameter passed in the `@MustCallChoice` position, and recurses on that parameter.  Let's rewrite the doc to be more like this, and leave out the details about handling receivers

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -428,26 +561,33 @@ private void handleSuccessorBlocks(
...
             Node last = nodes.get(nodes.size() - 1);
             CFStore storeAfter = typeFactory.getStoreAfter(last);
-            checkMustCall(assign, storeAfter, outOfScopeReason);
+            checkMustCall(setAssignCopy, storeAfter, outOfScopeReason);
           }
 
-          toRemove.add(assign);
+          toRemove.add(setAssignCopy);
+        } else {
+          setAssignCopy.removeIf(assign -> succRegularStore.getValue(assign.localVar) == null);
```
msridhar, line 586: This needs a comment indicating you are handling the case where _some_ vars go out of scope in the set.  Also, I think you can just make `setAssignCopy` in this else case rather than earlier, since this is the only place where mutation occurs.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -428,26 +561,33 @@ private void handleSuccessorBlocks(
...
+          defsCopy =
```
msridhar, line 587: Here I think a `remove(setAssign)` followed by `add(setAssignCopy)` would be clearer

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -514,40 +670,51 @@ private static boolean isTryWithResourcesVariable(LocalVariableNode lhs) {
...
-    List<String> mustCallValue = typeFactory.getMustCallValue(localVarWithTree.tree);
+      Set<LocalVarWithTree> localVarWithTreeSet, CFStore store, String outOfScopeReason) {
+
+    List<String> mustCallValue =
+        typeFactory.getMustCallValue(localVarWithTreeSet.iterator().next().tree);
     // optimization: if there are no must-call methods, we do not need to perform the check
     if (mustCallValue.isEmpty()) {
       return;
     }
     AnnotationMirror cmAnno;
```
msridhar, line 681: We should declare this variable in the loop below

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: Remove `defsCopy` also?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -514,40 +670,51 @@ private static boolean isTryWithResourcesVariable(LocalVariableNode lhs) {
...
 
-        checker.reportError(
-            localVarWithTree.tree,
-            "required.method.not.called",
-            formatMissingMustCallMethods(mustCallValue),
-            localVarWithTree.localVar.getType().toString(),
-            outOfScopeReason);
+      if (calledMethodsSatisfyMustCall(mustCallValue, cmAnno)) {
+        return;
+      } else if (!iter.hasNext()) {
```
msridhar, line 706: I don't understand this `!iter.hasNext()` check.  Do you only want to report an error if _all_ the variables in the set do not satisfy the must-call requirement?  Or are you trying to report the error on the final variable?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -428,26 +561,33 @@ private void handleSuccessorBlocks(
     }
     List<Node> nodes = block.getNodes();
     for (Block succ : getRelevantSuccessors(block)) {
-      Set<LocalVarWithTree> defsCopy = new HashSet<>(defs);
-      Set<LocalVarWithTree> toRemove = new HashSet<>();
+      Set<Set<LocalVarWithTree>> defsCopy = new LinkedHashSet<>(defs);
```
msridhar, line 564: These vars can both be `Set<ImmutableSet<LocalVarWithTree>>`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -605,11 +772,13 @@ static String formatMissingMustCallMethods(List<String> mustCallVal) {
    */
   private static class BlockWithLocals {
     public final Block block;
-    public final Set<LocalVarWithTree> localSetInfo;
+    public final ImmutableSet<ImmutableSet<LocalVarWithTree>> localSetInfo;
 
-    public BlockWithLocals(Block b, Set<LocalVarWithTree> ls) {
+    public BlockWithLocals(Block b, Set<Set<LocalVarWithTree>> ls) {
```
msridhar, line 777: The type of `ls` should be `Set<ImmutableSet<LocalVarWithTree>>`.  Then we can get rid of lines 779 and 780

msridhar (changes requested):
A few more comments

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +362,153 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+        LocalVarWithTree lhsLocalVarWithTreeNew =
+            new LocalVarWithTree(new LocalVariable((LocalVariableNode) lhs), node.getTree());
+        newDefs.add(ImmutableSet.of(lhsLocalVarWithTreeNew));
+        newDefs.remove(setContainingRhs);
+      }
+    }
+    return newDefs;
+  }
+
+  private Set<ImmutableSet<LocalVarWithTree>> updateDefs(
```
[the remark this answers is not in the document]
Nargeshdb: I wanted to keep the order of ImmutableSets in newDefs, but maybe it's not necessary. I'll update it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: I want to check the first case and also report the error for the first variable in the localVarWithTreeSet!

msridhar: I agree we want to check that all variables do not satisfy the must-call requirement before reporting an error.  We should restructure this code so that there is a boolean flag `mustCallSatisfied` that starts out `false`.  Then, if `calledMethodsSatisfyMustCall(mustCallValue, cmAnno)` returns true, set `mustCallSatisfied` to `true` and break out of the loop.  Then, after the loop, if `mustCallSatisfied` is `false`, report the error on the first variable in the set.  That will make the logic clearer.

object-construction-checker/tests/mustcall/MustCallChoiceImpl.java
```diff
@@ -0,0 +1,24 @@
+// A simple test that the extra obligations that MustCallChoice imposes are
+// respected.
+
+// @skip-test until the checks are implemented
```
msridhar, line 4: @Nargeshdb can you comment here on the unexpected error you got for this file, so @kelloggm can take a look?

object-construction-checker/tests/mustcall/MustCallChoiceOwningField.java
```diff
@@ -0,0 +1,28 @@
+// Based on a MustCallChoice scenario in Zookeeper.
+
+import org.checkerframework.checker.objectconstruction.qual.*;
+import org.checkerframework.checker.mustcall.qual.*;
+import org.checkerframework.checker.calledmethods.qual.*;
+import java.io.*;
+
+public @MustCall("shutdown") class MustCallChoiceOwningField {
```
msridhar, line 8: How is this testing `@MustCallChoice`?  I don't see that annotation in the test

Nargeshdb: It's written in stub files.

msridhar:
@kelloggm there are a couple of open comments but I think this is ready enough for you to take another look

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: I updated this code to use mustCallSatisfied boolean.

object-construction-checker/tests/socket/ACSocketTest.java
Nargeshdb: SelectionKey doesn't implement Closeable, so we can't handle @Owning annotation on that. I commented out this test case.

object-construction-checker/tests/mustcall/MustCallChoiceImpl.java
Nargeshdb: I uncommented it. the unexpected error that I got is documented in the test.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +358,144 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+    }
+
+    if (node instanceof MethodInvocationNode || node instanceof ObjectCreationNode) {
+      List<Node> arguments = getArgumentsOfMethodOrConstructor(node);
+      List<? extends VariableElement> formals = getFormalsOfMethodOrConstructor(node);
+
+      for (int i = 0; i < arguments.size(); i++) {
+        if (typeFactory
+                .getTypeFactoryOfSubchecker(MustCallChecker.class)
+                .getDeclAnnotationNoAliases(formals.get(i), MustCallChoice.class)
```
kelloggm, line 454: I think this should be a call to `ObjectConstructionAnnotatedTypeFactory#hasMustCallChoice` rather than a direct check via the must-call checker

object-construction-checker/tests/mustcall/MustCallChoiceLayeredStreams.java
```diff
@@ -0,0 +1,19 @@
...
+
+class MustCallChoiceLayeredStreams {
+    InputStream cache;
+    public InputStream createInputStream(String filename) throws FileNotFoundException {
+        if (cache == null) {
+            // The real version of this uses a mix of JDK and custom streams, so it makes more sense...
+            // TODO we shouldn't report a warning here and the code is okay because the cache is non-owning,
+            // and the caller of createInputStream is the owner of all of these streams.
+            // :: error: required.method.not.called
+            cache = new DataInputStream(new BufferedInputStream(new FileInputStream(new File(filename))));
```
kelloggm, line 15: The problem here is the dataflow through the non-owning field, right?

object-construction-checker/tests/socket/ACSocketTest.java
```diff
@@ -479,12 +477,19 @@ private SSLSocket configureSSLSocket(@Owning SSLSocket socket, boolean isClientS
 
 
     private void updateSocketAddresses( SelectionKey sockKey ) {
-        // error: required.method.not.called
         Socket socket = ((SocketChannel) sockKey.channel()).socket();
         SocketAddress localSocketAddress = socket.getLocalSocketAddress();
         SocketAddress remoteSocketAddress = socket.getRemoteSocketAddress();
     }
 
+    // SelectionKey doesn't implement Closeable
```
kelloggm, line 485: Rather than commenting this out, can we just write the correct annotation on `SelectionKey`? I'm not familiar with the API, so I'm not sure exactly what the spec is, but we ought to try to fix it if we really should report an error here.

object-construction-checker/tests/socket/InputOutputStreams.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    // getInputStream()/getOutputStream() can throw IOException in three different scenarios:
+    // 1) The underlying socket is already closed
+    // 2) The underlying socket is not connected
+    // 3) The underlysing socket input is shutdown
+    // In the first case our checker always reports a false positive but for the second case
+    // and third case our checker has to verify that close is called on the underlying resource.
+    // So, because sock.getInputStream() can throw IOException, "is" can be null, then sock will
+    // remain open. So, it's a true pasitive warning.
```
kelloggm, line 27: pasitive -> positive

object-construction-checker/tests/socket/OptionalSocket.java
```diff
@@ -0,0 +1,27 @@
...
+// These are sort-of a must call choice situation, but it's not as clear.
+// I suspect our MCC implementation won't actually handle these cases, in
+// which case it's fine with me to skip this test for now. - Martin
+
+import java.util.*;
+import java.net.*;
+import java.io.*;
+import org.checkerframework.checker.objectconstruction.qual.*;
+
+class OptionalSocket {
```
kelloggm, line 11: Rather than adding these expected errors, we should skip this test and keep it correct. To do that, add the following line at the beginning of the file:

```
// @skip-test
```

I did the same with the `MustCallChoice` tests in the MustCallChecker's test directory that require us to actually implement checks that MCC is used correctly.

kelloggm (changes requested):

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Socket.astub
```diff
@@ -77,10 +77,10 @@ class ServerSocket implements Closeable {
     InetAddress getInetAddress();
     int getLocalPort();
     SocketAddress getLocalSocketAddress();
-    @Owning Socket accept() throws IOException;
+    @MustCallChoice Socket accept(@MustCallChoice ServerSocket this) throws IOException;
```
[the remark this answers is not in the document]
Nargeshdb: [https://github.com/kelloggm/object-construction-checker/pull/242/commits/73c6d231dbb2fbab8081292b756750c8770dfc73](url)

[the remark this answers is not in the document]
Nargeshdb: [](https://github.com/kelloggm/object-construction-checker/pull/242/commits/73c6d231dbb2fbab8081292b756750c8770dfc73)

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +357,141 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+  }
+
+  /**
+   * This method tries to find a local variable passed as a @MustCallChoice parameter. In the base
+   * case, if {@code node} is a local variable, it just gets returned. Otherwise, if node is a call
+   * (or a call wrapped in a cast), the code finds the parameter passed in the @MustCallChoice
+   * position, and recurses on that parameter.
+   *
+   * @param node
+   * @return {@code node} iff {@code node} represents a local variable, otherwise null
```
msridhar, line 439: This documentation on the `@return` is not correct

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -289,20 +305,22 @@ private void doOwnershipTransferToParameters(Set<LocalVarWithTree> newDefs, Node
...
-            newDefs.remove(getAssignmentTreeOfVar(newDefs, local));
+            newDefs.remove(getSetContainingAssignmentTreeOfVar(newDefs, local));
           }
         }
       }
     }
   }
 
-  private void handleReturn(ReturnNode node, ControlFlowGraph cfg, Set<LocalVarWithTree> newDefs) {
+  private Set<ImmutableSet<LocalVarWithTree>> handleReturn(
```
msridhar, line 315: Can we just return `void` here now?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -324,11 +342,12 @@ private boolean isTransferOwnershipAtReturn(ControlFlowGraph cfg) {
     return false;
   }
 
-  private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs) {
+  private Set<ImmutableSet<LocalVarWithTree>> handleAssignment(
```
msridhar, line 345: Can we just return `void` here now?

msridhar (approved):
Approved assuming a couple more minor comments are addressed

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Socket.astub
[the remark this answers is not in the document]
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/242/commits/73c6d231dbb2fbab8081292b756750c8770dfc73

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -215,9 +227,25 @@ private boolean isOwningAssignmentLhs(Element elem) {
...
-  private boolean shouldSkipInvokePseudoAssignCheck(Tree callTree) {
+  private boolean shouldSkipInvokePseudoAssignCheck(
+      Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
+    Tree callTree = node.getTree();
+    List<String> mustCallVal = typeFactory.getMustCallValue(callTree);
+    if (mustCallVal.isEmpty()) {
+      return true;
+    }
+    // In this case, we are handling method invocation nodes that are not assigned to a local
+    // variable node but the receiver of the method is in the defs
```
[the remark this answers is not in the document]
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/242/commits/73c6d231dbb2fbab8081292b756750c8770dfc73

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +358,144 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+   * This method tries to find a local variable passed as a @MustCallChoice parameter. In the base
+   * case, if {@code node} is a local variable, it just gets returned. Otherwise, if node is a call
+   * (or a call wrapped in a cast), the code finds the parameter passed in the @MustCallChoice
+   * position, and recurses on that parameter.
+   *
+   * @param node
+   * @return {@code node} iff {@code node} represents a local variable, otherwise null
+   */
+  private @Nullable LocalVariableNode getLocalPassedAsMustCallChoiceParam(Node node) {
+    if (node instanceof TypeCastNode) {
```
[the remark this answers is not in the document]
Nargeshdb: you're right. It's better to write a `while` loop. I added a test for that.
https://github.com/kelloggm/object-construction-checker/pull/242/commits/13a87b4e68738dc38e2a81caba9653248a8bcc9f

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -338,44 +358,144 @@ private void handleAssignment(AssignmentNode node, Set<LocalVarWithTree> newDefs
...
+      node = ((TypeCastNode) node).getOperand();
+    }
+
+    if (node instanceof MethodInvocationNode || node instanceof ObjectCreationNode) {
+      List<Node> arguments = getArgumentsOfMethodOrConstructor(node);
```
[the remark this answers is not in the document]
Nargeshdb: Yes, it's better to check this condition before the loop. Thanks.
https://github.com/kelloggm/object-construction-checker/pull/242/commits/8c05c12095b5a41bb50b62b154eacd0bb6d2dc87

object-construction-checker/tests/mustcall/MustCallChoiceLayeredStreams.java
Nargeshdb: We don't track non-owning fields. If it was non-final owning field, then the problem was this.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/242/commits/73a5d0d6f607756067b47b9b65bd846ff1a9c71d

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: Yes, we don't need to store the return value anymore.
https://github.com/kelloggm/object-construction-checker/pull/242/commits/73a5d0d6f607756067b47b9b65bd846ff1a9c71d

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/242/commits/73a5d0d6f607756067b47b9b65bd846ff1a9c71d

object-construction-checker/tests/socket/ACSocketTest.java
```diff
@@ -479,12 +477,28 @@ private SSLSocket configureSSLSocket(@Owning SSLSocket socket, boolean isClientS
 
 
     private void updateSocketAddresses( SelectionKey sockKey ) {
-        // error: required.method.not.called
```
msridhar, line 482: Just so we don't forget, after chatting this morning @Nargeshdb and I think there should still be a warning here, so we have a bug somewhere

object-construction-checker/tests/socket/ACSocketTest.java
msridhar: I don't think we can handle `SelectionKey` yet.  It has an internal pointer to a channel that a client can access by calling `channel()`.  That channel is not new, but one registered with some `Selector` in some other code.  We don't have a way of capturing that aliasing relationship with our annotations

object-construction-checker/tests/socket/ACSocketTest.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/242/commits/0aff468930f00936d5c80c5f307eddc1fff93edb

msridhar:
@kelloggm I think this is ready for another look from you

object-construction-checker/tests/socket/ACSocketTest.java
kelloggm: The checker does issue the error correctly on this code, so I'm going to uncomment it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -213,11 +219,32 @@ private boolean isOwningAssignmentLhs(Element elem) {
...
+    if (mustCallVal.isEmpty()) {
+      return true;
+    }
+    // In this case, we are handling method invocation nodes that are not assigned to a local
+    // variable node but the receiver of the method is in the defs
+    if (callTree.getKind() == Tree.Kind.METHOD_INVOCATION
+        || callTree.getKind() == Tree.Kind.NEW_CLASS) {
+      LocalVariableNode mustCallChoiceParam = getLocalPassedAsMustCallChoiceParam(node);
+      if (mustCallChoiceParam != null) {
+        if (isVarInDefs(defs, mustCallChoiceParam)) {
```
kelloggm, line 238: These lines could be simplified to `return isVarInDefs(defs, mustCallChoiceParam)`

kelloggm (approved):
one minor code improvement suggestion, but otherwise LGTM

PR #249, opened 2020-12-07
Must call count

Nargeshdb:
It reports the total number of resources and the number of resources handled correctly by the enclosing method.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -160,6 +159,7 @@ private void checkPseudoAssignToOwning(Node node) {
     if (mustCallVal.isEmpty()) {
       return;
     }
+    increaseNumMustCall();
```
msridhar, line 162: Please add a comment explaining why we choose to increase the count here

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -611,6 +612,7 @@ private void handleSuccessorBlocks(
           Set<LocalVarWithTree> setOfLocals = new LinkedHashSet<>();
           setOfLocals.add(new LocalVarWithTree(new LocalVariable(paramElement), param));
           init.add(ImmutableSet.copyOf(setOfLocals));
+          increaseNumMustCall();
```
msridhar, line 615: Why do we also increment here?  Please add a comment

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -192,6 +192,7 @@ private void checkPseudoAssignToOwning(Node node) {
       AnnotationMirror cmAnno =
           typeFactory.getAnnotatedType(tree).getAnnotationInHierarchy(typeFactory.top);
       if (!calledMethodsSatisfyMustCall(mustCallVal, cmAnno)) {
+        checker.numMustCallFailed++;
```
kelloggm, line 195: Rather than adding this before every call to `checker.reportError`, we could override `reportError` in `ObjectConstructionChecker` and add to this variable if the warning string is `"required.method.not.called"`.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -23,10 +27,21 @@
...
 
+  public static final String COUNT_MUST_CALL = "countMustCall";
+
+  /**
+   * The number of expressions with must-call obligations that were checked. Incremented only if the
+   * {@link #COUNT_MUST_CALL} option was supplied.
+   */
+  int numMustCall = 0;
+
+  int numMustCallFailed = 0;
```
msridhar, line 43: Do we really need to track this explicitly?  Is this number just equal to the number of errors reported?  If so, it may be more robust to count the # of errors with a script after the fact.  With this approach, we need to encapsulate all error reporting in a single routine, so we don't forget to increment the field.  But even with that, we need to remember to always call our own error-reporting routine, which is annoying.

@kelloggm thoughts?

msridhar (changes requested):

kelloggm (approved):
Overall this looks good, and my informal testing matches my expectations. Thanks for figuring out how to get this working!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
kelloggm: I agree with you. I think we should override `BaseTypeChecker#reportError` and increment this field there whenever we issue a `required.methods.not.called` error.

(I suggested this in another comment, which it seems I posted about 30 seconds later than you posted this!)

msridhar: Ok sounds good.  Let's go with overriding `BaseTypeChecker#reportError`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: Actually, when I wanted to add a comment I noticed that here is not the right place to increment NumMustCall because checkPseudoAssignToOwning() is called not only in handleInvocation(), but also in handleTypeCast(). So, I updated the code to call incrementNumMustCall() only in handleInvocation() where the node is creating a new resource that is not skipped.
https://github.com/kelloggm/object-construction-checker/pull/249/commits/5260a7c49c6dc2c042e59816683cc968c422883c

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/249/commits/5260a7c49c6dc2c042e59816683cc968c422883c

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/249/commits/5260a7c49c6dc2c042e59816683cc968c422883c

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/249/commits/5260a7c49c6dc2c042e59816683cc968c422883c

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -611,6 +612,8 @@ private void handleSuccessorBlocks(
           Set<LocalVarWithTree> setOfLocals = new LinkedHashSet<>();
           setOfLocals.add(new LocalVarWithTree(new LocalVariable(paramElement), param));
           init.add(ImmutableSet.copyOf(setOfLocals));
+          // Increment numMustCall for each @Owning parameter tracked by the enclosing method
+          incrementNumMustCall();
```
msridhar, line 616: 👍

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -169,9 +173,7 @@ private void checkPseudoAssignToOwning(Node node) {
         assignedToOwning = isOwningAssignmentLhs(elementForType);
       } else if (assignmentContext instanceof MethodParameterContext) {
         // must be an @Owning or @MustCallChoice parameter
-        assignedToOwning =
-            typeFactory.getDeclAnnotation(elementForType, Owning.class) != null
-                || typeFactory.hasMustCallChoice(elementForType);
+        assignedToOwning = typeFactory.getDeclAnnotation(elementForType, Owning.class) != null;
```
msridhar, line 176: Why do we have this logic change (removing the or)?

msridhar (approved):
Looks good, just had one question (not blocking)

PR #252, opened 2020-12-08
Add command-line option to disable UnconnectedSocketChecker

Nargeshdb:


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -27,6 +27,8 @@
 
   public static final String CHECK_MUST_CALL = "checkMustCall";
 
+  public static final String Disable_Unconnected_Socket = "disableUnConnectedSock";
```
msridhar, line 30: Rename to use all caps:

```suggestion
  public static final String DISABLE_UNCONNECTED_SOCKET = "disableUnConnectedSock";
```

msridhar: You also need to actually not add the `UnconnectedSocketChecker` sub-checker inside `getImmediateSubcheckerClasses()`

msridhar (changes requested):

msridhar:
Also need to base changes on latest `always-call-checker` branch to avoid conflicts

msridhar (approved):

PR #253, opened 2020-12-09
Handle non-source code differently when checking pseudo-assignments

kelloggm:
This prevents warnings from being issued in non-source locations, which are really hard to interpret. Without this, the new case case issues a `required.method.not.called` error for the entire class due to the foreach loop, which is useless to the user.

I don't think there are any generated elements that won't be assigned into a local variable, but I might not be remembering something.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -115,7 +115,6 @@ class MustCallInvokedChecker {
           new LinkedHashSet<>(curBlockLocals.localSetInfo);
 
       for (Node node : nodes) {
-
         if (node instanceof AssignmentNode) {
           handleAssignment((AssignmentNode) node, newDefs);
         } else if (node instanceof ReturnNode) {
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -114,7 +114,6 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
           new LinkedHashSet<>(curBlockLocals.localSetInfo);
 
       for (Node node : nodes) {
-
```
msridhar, line 117: remove whitespace change?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -717,7 +718,6 @@ private void checkMustCall(
           .noneMatch(localVarTree -> localVarWithTreeSet.contains(localVarTree))) {
         LocalVarWithTree firstlocalVarWithTree = localVarWithTreeSet.iterator().next();
         reportedMustCallErrors.add(firstlocalVarWithTree);
-
```
msridhar, line 720: remove whitespace change?

msridhar (approved):
LGTM and makes sense.  Couple optional nits

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: Sorry, didn't notice I had made these (I tried a few different solutions for this before I settled on the one in the PR...).

I kind of prefer the whitespace changes, but I'm happy to revert them if you feel strongly.

msridhar: Keeping is ok with me

PR #254, opened 2020-12-09
Use input block store rather than successor block store when checking conditional blocks

kelloggm:
Fixes an extremely weird error I was getting in Zookeeper, from which the test case in this PR is derived. Check out the version of the test in https://github.com/kelloggm/object-construction-checker/commit/1aeb21983e198c73489f99f0d266a4b5e6232561 for a failing version that's a lot closer to the original Zookeeper code.

As far as I can tell, the issue here is that the successor to the empty conditional block doesn't have a type for `fis1`. Why this only comes up when you have two nested conditionals, I'm not sure.

I considered using the original code for special blocks and creating a separate code path for conditionals, but all our tests are passing with this change, which is simpler, so I went with this one. If a reviewer can write a test that fails because of this change that doesn't use conditionals, I'd love to see it.


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -573,8 +573,7 @@ private void handleSuccessorBlocks(
             || setAssign.stream()
                 .allMatch(assign -> succRegularStore.getValue(assign.localVar) == null)) {
           if (nodes.size() == 0) { // If the cur block is special or conditional block
```
msridhar, line 575: I think we should only enter this condition when the current block is a conditional block.  Could we add an assertion to this effect?  We should not be doing analysis within special blocks, I think.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -573,8 +573,7 @@ private void handleSuccessorBlocks(
...
-            checkMustCall(setAssign, succRegularStore, reasonForSucc);
-
+            checkMustCall(setAssign, analysis.getInput(block).getRegularStore(), reasonForSucc);
```
msridhar, line 576: This change makes sense to me in the case where none of the local variables in `setAssign` appear in `succRegularStore`.  In such a case, the variable is going out of scope, and it doesn't make sense to pass `succRegularStore` to `checkMustCall`.

Now let's consider the case where some local variable _does_ appear in `succRegularStore`.  There, we have the question of whether some variable type in `succRegularStore` could differ from its type in `analysis.getInput(block).getRegularStore()`.  But since we know `nodes.size() == 0`, I _think_ this is not possible.

At the least, can we add a comment explaining why we are using the block we choose here, maybe with some of the argument above?  We could also add some logic to only pass `analysis.getInput(block).getRegularStore()` when the `setAssign.stream().allMatch(...)` expression above is `true`.  But I don't think that should be necessary.

msridhar (changes requested):
I think the change is correct.  A bit of feedback

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: Adding the following here immediately causes our tests to fail:

```
            if (block.getType() != BlockType.CONDITIONAL_BLOCK) {
              throw new BugInCF("unexpected non-conditional block: " + block.getType());
            }
```

The throw exception looks like this:

```
    :-1: other: error: unexpected non-conditional block: SPECIAL_BLOCK
      ; The Checker Framework crashed.  Please report the crash.
      Compilation unit: /Users/kelloggm/Research/object-construction-checker/object-construction-checker/tests/mustcall/ACOwning.java
      Last visited tree at line 5 column 1:
      class ACOwning {
      Exception: java.lang.Throwable; java.lang.Throwable
        at org.checkerframework.javacutil.BugInCF.<init>(BugInCF.java:16)
        at org.checkerframework.checker.objectconstruction.MustCallInvokedChecker.handleSuccessorBlocks(MustCallInvokedChecker.java:578)
...
```

kelloggm: (There is also at least one crash in the `socket` tests.)

msridhar: Ok, let's leave that assertion out then.  I guess we do propagate to special blocks.

kelloggm:
@msridhar coincidentally, I've found the test case that this test breaks. I just committed in #255 - it's the one that uses `@EnsuresCalledMethodsIf`. In that case, the conditional block's store doesn't contain the `@CalledMethods("close")` type -only the successor block does. I'm not sure why, but this throws a wrench in this PR.

I've also written the comment you asked for, and I'll commit that and the merge while I investigate.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: > We could also add some logic to only pass analysis.getInput(block).getRegularStore() when the setAssign.stream().allMatch(...) expression above is true.

Doing this fixes the problem related to `EnsuresCalledMethodsIf` I described below. I've pushed that version of the code.

msridhar (approved):

PR #256, opened 2020-12-13
Handle ternary expressions

msridhar:
At a high level, the handling is similar to type casts.  Handling ternary expressions on the RHS of an assignment to some local variable `x` requires some care.  If an operand is a method call, we can treat it as if it were just normally being assigned to `x`, since whenever that expression branch executes, the result of the call is assigned to `x`.  But if an operand is a local variable `y`, we should _not_ do ownership transfer, as that branch of the expression is only executed conditionally and it is possible that `y` is not copied to `x`.

There is also some refactored code to handle ternary expressions and casts nested within each other; deeply-nested casts led to false positives before.  See changes to `ACRegularExitPoint` for the new tests.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -5,6 +5,7 @@ import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
 import com.sun.tools.javac.code.Type;
+
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.HashSet;
@@ -44,6 +45,7 @@ import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
 import org.checkerframework.dataflow.cfg.node.Node;
 import org.checkerframework.dataflow.cfg.node.ObjectCreationNode;
 import org.checkerframework.dataflow.cfg.node.ReturnNode;
+import org.checkerframework.dataflow.cfg.node.TernaryExpressionNode;
 import org.checkerframework.dataflow.cfg.node.TypeCastNode;
 import org.checkerframework.dataflow.expression.LocalVariable;
 import org.checkerframework.framework.flow.CFAnalysis;
@@ -143,7 +145,7 @@ class MustCallInvokedChecker {
     doOwnershipTransferToParameters(newDefs, node);
     // If the method call is nested in a type cast, we won't have a proper AssignmentContext for
     // checking.  So we defer the check to the corresponding TypeCastNode
-    if (!nestedInTypeCast(node) && !shouldSkipInvokePseudoAssignCheck(node, newDefs)) {
+    if (!nestedInCastOrTernary(node) && !shouldSkipInvokePseudoAssignCheck(node, newDefs)) {
       // If the node is not skipped by the shouldSkipInvokePseudoAssignCheck() it means we have a
       // method invocation or object creation node that creates a new resource, so we increment
       // the numMustCall
@@ -259,10 +261,10 @@ class MustCallInvokedChecker {
   }
 
   /**
-   * Checks if {@code node} is an invocation nested inside a TypeCastNode, by looking at the
-   * successor block in the CFG
+   * Checks if {@code node} is an invocation nested inside a {@link TypeCastNode} or a {@link
+   * TernaryExpressionNode}, by looking at the successor block in the CFG
    */
-  private boolean nestedInTypeCast(Node node) {
+  private boolean nestedInCastOrTernary(Node node) {
     if (!(node instanceof MethodInvocationNode || node instanceof ObjectCreationNode)) {
       throw new BugInCF("unexpected node type " + node.getClass());
     }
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -345,13 +378,68 @@ private boolean isTransferOwnershipAtReturn(ControlFlowGraph cfg) {
...
+      AssignmentNode node,
+      Set<ImmutableSet<LocalVarWithTree>> newDefs,
+      TernaryExpressionNode rhsTernary) {
+    // If handling the then operand of the ternary expression results in adding the assignment to
+    // newDefs, we do not want to also handle the else operand.  Otherwise, for an assignment like
+    // x = b ? new Foo() : new Foo();, the algorithm would add the assignment for the then case, but
+    // then report a false positive error for the else case, since based on the updated state of
+    // newDefs it looks like a tracked variable x is being overwritten.  To avoid these false
+    // positives, we track the size of newDefs before and after handling the then operand, and only
+    // recurse to the else operand if the size does not change.
```
kelloggm, line 408: I'm concerned about this, because it's possible that the then and else branches will have different, non-empty must-call obligations. Consider e.g.:

```
@MustCall("toString") class Foo extends Object { }
@MustCall("clone") class Bar extends Object { }

static void test_ternary(boolean b) {
    // :: error: required.method.not.called
    Object obj = b ? new Foo() : new Bar();
    // at this point, for soundness, we should be responsible for calling both toString and clone on obj...
    obj.toString();
}
```

Hopefully I'm wrong, but I'm concerned that this code won't handle this correctly, because it'll add the `toString` obligation and then not check the else expression.

kelloggm (approved):
This looks great - thanks!

I have one concern related to possibly-different must-call types in the two branches of a ternary. If that test passes, then we can merge this right away.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: This is a valid concern; will look into it

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -121,29 +123,52 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
-    if (operand instanceof MethodInvocationNode || operand instanceof ObjectCreationNode) {
-      if (!shouldSkipInvokePseudoAssignCheck(operand, defs)) {
-        checkPseudoAssignToOwning(node);
-      }
+  private void handleTypeCastOrTernary(Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
+    if (!(node instanceof TypeCastNode || node instanceof TernaryExpressionNode)) {
+      throw new BugInCF("unexpected node type " + node.getClass());
+    }
+    if (!nestedInCastOrTernary(node) && hasNestedPseudoAssignInvoke(node, defs)) {
+      checkPseudoAssignToOwning(node);
```
Nargeshdb, line 140: Can we replace hasNestedPseudoAssignInvoke() with !shouldSkipInvokePseudoAssignCheck()? I can't understand why we need to check TypeCastNode and TernaryExpressionNode in hasNestedPseudoAssignInvoke() again.

Nargeshdb (changes requested):

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: I added the test in 4c3789d, and it works.  The reason it works is that the `@MustCall` type for the local variable is computed by the Must Call Checker, which correctly computes the LUB for the ternary expression as `@MustCall({"toString","clone"})`.  So `MustCallInvokedChecker` indeed finds the issue.

I think there is a way to clean up the above code more, by just storing the relevant (sub-)expression on the RHS of an assignment in the `LocalVarWithTree` object rather than the whole assignment.  Then we might be able to get rid of all the counting logic.  But that is a broader change we can look into in a follow-up PR.  I'll open an issue on that once this lands.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: @Nargeshdb my understanding was that `shouldSkipInvokePseudoAssignCheck()` is intended to work for nodes that are invocations.  Here, `node` is either a `TypeCastNode` or a `TernaryExpressionNode`.  So, `hasNestedPseudoAssignInvoke()` does the work of recursing through the operands of `node` looking for invocation nodes, and then passing those nodes to `shouldSkipInvokePseudoAssignCheck()`.  If _any_ of the nested invokes requires a pseudo-assign check, `hasNestedPseudoAssignInvoke()` will return true.  Does that sound right?  We do not want to do a pseudo-assign check if there are no nested invokes.  E.g., we don't want to do the check for `b ? 3 : -3`.

Nargeshdb (approved):

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: In 30b5eae I added a test that shows the need for `hasNestedPseudoAssignInvoke()`

PR #261, opened 2020-12-16
check if the receiver parameter of @MustCallChoice is a resouce varia…

Nargeshdb:


msridhar:
Rather than having to sprinkle these checks in everywhere, I wonder if it would be cleaner to have the Must Call Checker remove `close` from the `@MustCall` type of any try-with-resources variable.  @kelloggm do you think that would be doable?

mernst:
> Rather than having to sprinkle these checks in everywhere, I wonder if it would be cleaner to have the Must Call Checker remove `close` from the `@MustCall` type of any try-with-resources variable. @kelloggm do you think that would be doable?

I agree with this proposal.  (That's how I assumed you were planning to handle it.)

kelloggm:
Yes, that's do-able. I think I could implement it that way pretty easily. I'll make a PR when I get a chance.

kelloggm:
@msridhar @Nargeshdb I've implemented what Manu suggested in the MustCall Checker. Please have a look.

The tests are still failing. I think that the MC checker is doing its job here, and there is something wrong with both of the test methods:

1. the first error is caused by the file descriptor problem we discussed earlier. We should probably remove this test from this PR, because it'll need to be addressed separately (e.g. in #262). The error is:
    ```  
    object-construction-checker/tests/socket/TryWithResourcesSimple.java:35: error: [required.method.not.called] @MustCall method(s) close,  for variable/expression not invoked.  The type of object is: java.nio.channels.FileChannel.  Reason for going out of scope: never assigned to an @Owning location
            file.setLength(outSpecial.getChannel().position());
    ```

2. the second error is caused by use not having a stub file for `RandomAccessFile` with an appropriate `@MustCallChoice` annotation:
    ```
    object-construction-checker/tests/socket/TryWithResourcesSimple.java:35: error: [required.method.not.called] @MustCall method(s) close,  for variable/expression not invoked.  The type of object is: java.nio.channels.FileChannel.  Reason for going out of scope: never assigned to an @Owning location
            file.setLength(outSpecial.getChannel().position());
    ```
    A similar test using `Socket` and `getInputStream` works in the Must Call Checker tests, so this should too once we have that stub.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -94,6 +96,9 @@ protected void addComputedTypeAnnotations(Tree tree, AnnotatedTypeMirror type, b
     if (TypesUtils.isPrimitiveOrBoxed(type.getUnderlyingType())) {
       type.replaceAnnotation(BOTTOM);
     }
+    if (isDeclaredInTryWithResources(TreeUtils.elementFromTree(tree))) {
+      type.replaceAnnotation(BOTTOM);
```
msridhar, line 100: Is it possible to define a class that implements `Closeable` and also has additional `@MustCall` requirements?  Admittedly a corner case, but if it's possible, we may want to tweak this logic to not set the type to bottom, but instead to just remove the `close` method from the type.  Thoughts?

kelloggm: Ah - you're right. I'll fix this.

kelloggm: @msridhar I've rewritten to handle this case, and I added a stub for one of the failures in the tests. If we remove the other test, we can merge this if you all are okay with it.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
```diff
@@ -71,6 +75,21 @@ protected boolean skipReceiverSubtypeCheck(
...
+  /**
+   * Skip assignment checks for try-with-resources variables, because they're always @MustCall({}).
+   */
+  @Override
+  protected void commonAssignmentCheck(
+      Tree varTree,
+      ExpressionTree valueExp,
+      @CompilerMessageKey String errorKey,
+      Object... extraArgs) {
+    if (TreeUtils.elementFromTree(varTree).getKind() == ElementKind.RESOURCE_VARIABLE) {
```
msridhar, line 87: @kelloggm do we still need this change?  If we do, we should update the Javadoc as it's out of date.

msridhar (changes requested):
I removed the failing test case.  One more comment

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
kelloggm: Nice catch - I didn't think about this when I made the other change last night. This is trickier.

The reason we need this at all is that the RHS of assignments in the `try` block still might have "close" in their MC type. For example, in this code:
```
try (Socket s = new Socket(addr, port)) { } catch (Exception e) { }
```
`Socket s` (the LHS) is correctly assigned the type `@MustCall({})` by taking it's normal type and removing "close", because it's a try-with-resources variable. The RHS, however, has no indication that it's a try-with-resources variable; it's such a `new Socket` - which correctly has the type `@MustCall("close")`. Without this change, then we issue an incompatible assignment warning here.

I think what we actually need to check here is that if the LHS is a resource variable, then we should remove `"close"` from the RHS type and then do the check. I'll try to get that working.

msridhar: Thanks for the explanation.  Based on what you wrote, I am not sure if you need to make a further change here, other than updating the comment.  Can you give a test case where the initial assignment for a try-with-resources variable _should_ yield an error?  I would hope these variables must be (effectively) final, so they can't be re-assigned.

kelloggm: > I would hope these variables must be (effectively) final, so they can't be re-assigned.

That's correct - the compiler inserts the `final` keyword IIRC.

> Can you give a test case where the initial assignment for a try-with-resources variable should yield an error?

What about this:
```
@MustCall({"close", "myMethod"})
class Foo extends Socket { void myMethod() {} /* also constructors etc */ }
try (Socket s = new Foo(addr, port)) { } catch (Exception e) { }
```

I'm actually not sure how much we care about a case like this, but I think at this point we should warn, because the LHS isn't `@MustCall("myMethod")`.

I think the fix is actually to do the subtyping check without close on both sides in the resource variable case. I think I can do that.

msridhar: I think it's fine to report an error at the initial assignment in this case for now.  Eventually I think the "right" fix would be to have the inferred type of `s` to be `@MustCall("myMethod")` and then to get rid of any special-casing of try-with-resources variables in `MustCallInvokedChecker`.  Then if `s.myMethod()` is invoked within the try block, the code will type check.

kelloggm: The current code won't report an error.

I agree that's what the code *should* look like, and I think right now this will work correctly if the declared type of the resource variable is `Foo` rather than `Socket`.

msridhar (approved):

kelloggm:
@mernst, if you don't mind taking a look at these changes, that would be helpful. I don't think it needs to block merging, though, if @msridhar needs this merged ASAP.

kelloggm:
Merging because I want to re-run Zookeeper with our latest changes.

PR #263, opened 2020-12-17
adding annotation in the stub files

Nargeshdb:


msridhar (approved):

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
```diff
@@ -17,13 +17,13 @@ class FileChannel {
 
 class SocketChannel extends AbstractSelectableChannel implements ByteChannel, ScatteringByteChannel, GatheringByteChannel, NetworkChannel {
     protected SocketChannel(SelectorProvider arg0);
-    static SocketChannel open() throws IOException;
-    static SocketChannel open(SocketAddress arg0) throws IOException;
+    @MustCallChoice static SocketChannel open(@MustCallChoice SocketChannel this) throws IOException;
+    @MustCallChoice static SocketChannel open(@MustCallChoice SocketChannel this, SocketAddress arg0) throws IOException;
```
kelloggm, line 21: These annotations aren't correct, and produce stub file parse errors.

These static methods open new `SocketChannel`s - the former opens an `@Unconnected` socket and the latter opens a connected socket (the unconnected socket checker has a stub for them already).

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
```diff
@@ -41,13 +41,13 @@ class SocketChannel extends AbstractSelectableChannel implements ByteChannel, Sc
 
 class ServerSocketChannel extends AbstractSelectableChannel implements NetworkChannel {
     protected ServerSocketChannel(SelectorProvider arg0);
-    static ServerSocketChannel open() throws IOException;
+    @MustCallChoice static ServerSocketChannel open(@MustCallChoice ServerSocketChannel this) throws IOException;
```
kelloggm, line 44: Similar note to the above - this method is definitely not MCC

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
```diff
@@ -41,13 +41,13 @@ class SocketChannel extends AbstractSelectableChannel implements ByteChannel, Sc
...
     int validOps();
-    ServerSocketChannel bind(SocketAddress arg0) throws IOException;
-    ServerSocketChannel bind(SocketAddress arg0, int arg1) throws IOException;
+    @MustCallChoice ServerSocketChannel bind(@MustCallChoice ServerSocketChannel this, SocketAddress arg0) throws IOException;
+    @MustCallChoice ServerSocketChannel bind(@MustCallChoice ServerSocketChannel this, SocketAddress arg0, int arg1) throws IOException;
```
kelloggm, line 47: I think these are correct - the Javadoc for `bind` says that it returns the same channel it's called on.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
```diff
@@ -41,13 +41,13 @@ class SocketChannel extends AbstractSelectableChannel implements ByteChannel, Sc
...
     <T> ServerSocketChannel setOption(SocketOption<T> arg0, T arg1) throws IOException;
     @MustCallChoice ServerSocket socket(@MustCallChoice ServerSocketChannel this);
-    SocketChannel accept() throws IOException;
+    @MustCallChoice SocketChannel accept(@MustCallChoice ServerSocketChannel this) throws IOException;
```
kelloggm, line 50: This annotation is wrong - `accept` produces a new socket. See e.g. https://docs.oracle.com/javase/7/docs/api/java/nio/channels/ServerSocketChannel.html#accept().

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -132,6 +113,34 @@ class BufferedInputStream extends FilterInputStream {
     void close() throws IOException;
 }
 
+class FileOutputStream extends OutputStream {
+    FileOutputStream(String arg0) throws FileNotFoundException;
```
kelloggm, line 117: In general in the future, can we avoid adding an entry for every method in a class? It makes it difficult to find what's actually annotated.

kelloggm (commented):
This is already merged, but I think it's incorrect. I noticed these annotations while trying to fix the file descriptor problem, because I was running with `-AstubDebug`, which flagged them.

@msridhar @Nargeshdb

msridhar:
@Nargeshdb can you file a follow-up PR fixing these issues?  And yes, let's try to check these changes with `-AstubDebug` in the future.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
Nargeshdb: You're right, open will create a new socket from the client side.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
Nargeshdb: You're right, `accept()` will create a new socket from the client-side.

PR #264, opened 2020-12-17
FileDescriptor problem

kelloggm:
@msridhar @Nargeshdb I looked into the issue with `FileDescriptor`s we discussed on our call yesterday. Thanks to @Nargeshdb for putting together a test case.

I wrote stubs that I think are correct wrt FDs, and included them here. The MC checker is handling them correctly (see the test case I wrote for it) as far as I can tell.

That said, the original test that @Nargeshdb wrote is still failing. I believe the reason isn't related to there being a `FileDescriptor` - see the version of the test that I added that only uses `Closeable` objects, which also fails.

The failure looks like this:
```
object-construction-checker/tests/socket/FileDescriptorTest.java:10: error: [required.method.not.called] @MustCall method(s) close,  for variable/expression not invoked.  The type of object is: java.io.RandomAccessFile.  Reason for going out of scope: possible exceptional exit due to in.close() with exception type java.io.IOException
        RandomAccessFile file = new RandomAccessFile(from, "rws");
                         ^
```

Note that this is due to the possible exceptional exit from calling `close` on `in`, which IIRC you all told me should be ignored.

Thoughts?

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
@@ -26,6 +26,7 @@ class DataOutputStream extends FilterOutputStream implements DataOutput {
 
 class RandomAccessFile {
     @MustCallChoice FileChannel getChannel(@MustCallChoice RandomAccessFile this);
+    @MustCallChoice FileDescriptor getFD(@MustCallChoice RandomAccessFile this) throws IOException;
 }
 
 class OutputStream { }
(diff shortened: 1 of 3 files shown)


msridhar:
Thanks a lot for this @kelloggm.  I agree (particularly based on your second test) that we have a separate bug here, unrelated to file descriptors.

On the positive side, very cool that we seem to be able to use `@MustCallChoice` to track file descriptors!

msridhar:
Oof, actually I see what is going on here I think.

https://github.com/kelloggm/object-construction-checker/blob/ce5c5b11124fccd02fc5817f1b584f3c0bd3d52f/object-construction-checker/tests/socket/FileDescriptorTest.java#L24-L32

The problem is the case where `getInputStream()` throws an `IOException`.  In that case, our analysis does not track the fact that `in` must be `null`.  So, the analysis thinks we can enter the then branch if the `if (in != null)` conditional, leading to `close()` throwing an exception and the Socket not getting closed.  This is a case where we need to track some nullness facts in a path-sensitive manner to get rid of the false positive.

@Nargeshdb maybe we can write a file-descriptor test somehow that avoids this exception + null tracking issue, just to make sure it works?

Nargeshdb:
> Oof, actually I see what is going on here I think.
> 
> https://github.com/kelloggm/object-construction-checker/blob/ce5c5b11124fccd02fc5817f1b584f3c0bd3d52f/object-construction-checker/tests/socket/FileDescriptorTest.java#L24-L32
> 
> The problem is the case where `getInputStream()` throws an `IOException`. In that case, our analysis does not track the fact that `in` must be `null`. So, the analysis thinks we can enter the then branch if the `if (in != null)` conditional, leading to `close()` throwing an exception and the Socket not getting closed. This is a case where we need to track some nullness facts in a path-sensitive manner to get rid of the false positive.
> 
> @Nargeshdb maybe we can write a file-descriptor test somehow that avoids this exception + null tracking issue, just to make sure it works?

@kelloggm Thanks Martin for checking this issue. 
@msridhar I wrote the following example and MC checker handles it correctly:
```
   public static void readPropertiesFile(File from) throws IOException {
        RandomAccessFile file = new RandomAccessFile(from, "rws");
        FileInputStream in = null;
        try {
            in = new FileInputStream(file.getFD());
            in.close();
        } catch (IOException e){
            file.close();
        }
   }
```

msridhar:
Ok, great!  Maybe then we can include Narges's test case, get CI passing by skipping some tests, and merge these stub file updates?

kelloggm:
@msridhar CI should pass now.

msridhar (approved):

msridhar:
@kelloggm is the target branch correct, so we should merge into `fix-try-with-resources-bug`?

kelloggm:
Oops. I'll make a new PR that's targeted correctly. Sorry.

PR #265, opened 2020-12-18
Annotations stubs

Nargeshdb:


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -132,6 +113,34 @@ class BufferedInputStream extends FilterInputStream {
     void close() throws IOException;
 }
 
+class FileOutputStream extends OutputStream {
```
kelloggm, line 116: I think these stubs for `FileOutputStream`, `FileInputStream`, and `RandomAccessFile` conflict with #266, which has fewer extraneous methods and comments that explain why the constructors are treated differently.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -215,6 +268,17 @@ class DigestOutputStream extends FilterOutputStream {
     String toString();
 }
 
+class DigestInputStream extends FilterInputStream {
+    protected MessageDigest digest;
+    @MustCallChoice DigestInputStream(@MustCallChoice InputStream arg0, MessageDigest arg1);
+    MessageDigest getMessageDigest();
```
kelloggm, line 274: What's the purpose of including these unannotated methods?

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -229,3 +293,31 @@ class Scanner {
     @MustCallChoice Scanner(@MustCallChoice ReadableByteChannel source, String charsetName);
     @MustCall({}) Scanner(String source);
 }
+
+class Formatter implements Closeable, Flushable {
+    Formatter();
+    @PolyMustCall Formatter(@PolyMustCall Appendable arg0);
```
kelloggm, line 299: Why `@PolyMustCall` rather than `@MustCallChoice`? Can you add a comment about it?

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/Stream.astub
```diff
@@ -229,3 +293,31 @@ class Scanner {
...
+    Formatter(Locale arg0);
+    @PolyMustCall Formatter(@PolyMustCall Appendable arg0, Locale arg1);
+    Formatter(String arg0) throws FileNotFoundException;
```
kelloggm, line 302: More extraneous unannotated methods

kelloggm (changes requested):
These stubs need to be cleaned up before they're merged, and they also need to be reconciled with the stubs I proposed in #266, which I think are simpler and include more annotations (e.g. on `getFD()` in `FileInputStream`).

PR #270, opened 2020-12-22
remove the receiver parameter at a return node

Nargeshdb:


object-construction-checker/tests/socket/ACSocketTest.java
```diff
@@ -480,5 +480,11 @@ private void updateSocketAddressesWithOwning(@Owning SelectionKey sockKey ) {
     @NotOwning Socket getSocket(Socket s) {
         return s;
     }
+
+    private ServerSocket testMCC() throws IOException{
```
msridhar, line 484: Can we do a better name for this method?  Maybe `testMCCParamInReturn()`?

msridhar (approved):
Please address the comment and then this looks good

PR #277, opened 2021-01-05
Use github actions

kelloggm:


kelloggm:
This looks like it's working, but I figured another set of eyes would be helpful @msridhar

msridhar (approved):
👍

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,19 @@
+name: Java CI
+
+on: [push]
```
mernst, line 3: This should include pull-request as well.

kelloggm: Good idea, thanks

PR #278, opened 2021-01-05
add @NotOwning annotation for ServletResponse and remove MCC annotati…

Nargeshdb:
- https://docs.oracle.com/javaee/6/api/javax/servlet/ServletResponse.html
- https://stackoverflow.com/questions/1159168/should-one-call-close-on-httpservletresponse-getoutputstream-getwriter

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
@@ -17,8 +17,8 @@ class FileChannel {
 
 class SocketChannel extends AbstractSelectableChannel implements ByteChannel, ScatteringByteChannel, GatheringByteChannel, NetworkChannel {
     protected SocketChannel(SelectorProvider arg0);
-    @MustCallChoice static SocketChannel open(@MustCallChoice SocketChannel this) throws IOException;
-    @MustCallChoice static SocketChannel open(@MustCallChoice SocketChannel this, SocketAddress arg0) throws IOException;
+    static SocketChannel open() throws IOException;
+    static SocketChannel open(SocketAddress arg0) throws IOException;
     int validOps();
     SocketChannel bind(SocketAddress arg0) throws IOException;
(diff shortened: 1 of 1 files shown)


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
```diff
@@ -41,7 +41,7 @@ class SocketChannel extends AbstractSelectableChannel implements ByteChannel, Sc
 
 class ServerSocketChannel extends AbstractSelectableChannel implements NetworkChannel {
     protected ServerSocketChannel(SelectorProvider arg0);
-    @MustCallChoice static ServerSocketChannel open(@MustCallChoice ServerSocketChannel this) throws IOException;
+    static ServerSocketChannel open() throws IOException;
```
kelloggm, line 44: I thought we'd already done this, but it's nbd.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
```diff
@@ -61,3 +61,10 @@ class AbstractInterruptibleChannel {
     @EnsuresCalledMethodsIf(expression="this", result=false, methods={"close"})
     boolean isOpen(AbstractInterruptibleChannel this);
 }
+
+package javax.servlet;
+
+interface ServletResponse {
+    @NotOwning ServletOutputStream getOutputStream() throws IOException;
```
kelloggm, line 68: Can you add the stack overflow link in the PR description here? It justifies why these annotations are correct, and if we ever run into a situation where we've forgotten we'll look for justification here rather than in the PR description.

kelloggm (approved):
Minor comments, lgtm otherwise

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/NotOwning.astub
Nargeshdb: It was in a PR that was closed before being merged.

PR #280, opened 2021-01-05
Check non-final owning fields at re-assignments rather than always issuing a warning at their declaration

kelloggm:
This was motivated by our discussion yesterday about assignments to fields when the field is definitely null. We thought that the must-call checker might be missing some refinement logic, but this PR includes a test case to show why that isn't the case.

However, I looked at the code in the MC invoked checker and decided to try fixing it, and this seems to do the right thing.


msridhar:
Overall, this change looks great, and I'm really glad that this kind of checking can work!

Reviewing this PR, I had a separate soundness concern with non-final fields.  For the new test class `NonFinalFieldOnlyOverwrittenIfNull`, what if a client writes this code:

```java
NonFinalFieldOnlyOverwrittenIfNull n = new NonFinalFieldOnlyOverwrittenIfNull();
n.set("foo.txt");
n.close();
n.set("bar.txt");
```

I believe the above code will type check, but it will still have a leak.  The issue is we do not currently check that no further operations are performed on an object _after_ `close()` gets called.  Further, I don't believe that such a property is accumulation; we'd have to check for operations through any alias for soundness.

For final fields, this is not a concern, since the field cannot be re-assigned after the constructor.  So, for whatever object `o` gets assigned into the field, as long as the `@MustCall` method of the enclosing class gets called, `o`'s `@MustCall` method will definitely get called.

@kelloggm @Nargeshdb do you agree that the above is a problem?  Any thoughts on a solution?

kelloggm:
@msridhar your point is a good one.

I think with the current test case, this would be fine - the second call to `set` wouldn't create a new resource, because the internal resource is already non-null.

That said, I could easily rewrite the test case to expose the problem. All that's needed is to make `close` set the field to `null`:

```
@EnsuresCalledMethods(value="this.is", methods="close")
public void close() throws Exception {
    this.is.close();
    this.is = null;
}
```

In this case, you're right - given the same call sequence you described, then `is` will be set to null again when `set` is called a second time, so the call to `set` will create a new must call obligation.

One possible solution that immediately comes to mind is the `@AddsMustCall` declaration annotation we were discussing for unconnected sockets on Monday - if we required any method that sets a non-final, owning field to have one, then maybe this would be fine?

Except I'm not confident that it would, because that still leaves us with the problem of needing to "restart" the accumulation process after `close` is called - with our current analysis, we'll still analyze the client (the code in your example) and conclude that `n` is `@MustCall("close")` and `@CalledMethods("close")` - which match, so we'll permit it.

I'll need to think about this more.

msridhar:
I don't see an obvious way around the need to check that some method was _not_ called on an object here.  At the same time, it seems like a shame if we cannot do any useful checking for non-final fields.  It might be worth also considering some kind of unsound (off-by-default?) checking for non-final fields that gives errors for bad field overwrites, but does not warn about the case where an object is "revived" after is `@MustCall` method is invoked.

kelloggm:
(This is mostly a note to myself so I remember where I am on Monday. Others could treat it as a progress report if interested.)

The reseting of MustCalls portion of ResetMustCall (what I've decided to call the simpler version of CreateMustCall that I mentioned in my email) is implemented and appears to be working. The decision to change the name (and semantics) was motivated by the fact that what we really want for both the local variable cases and the unconnected socket cases we've seen is to reset the type of the target variable to the default type for its class. That's what I've actually implemented, and it makes the annotation simpler.

What I've implemented so far is that the annotation definition itself (including docs) and code in the Must Call Checker to change the type of a variable that's the target of one of these annotations to its default type when the method is invoked. See the changes in `MustCallTransfer.java`. I've also upgraded the project to CF 3.9.0, because of the changes to `Receiver`/`JavaExpression`, which are heavily used here. I also deleted the unconnected socket checker rather than upgrading it, since this change will obsolete it anyway.

Remaining work items:

- [ ] In the Object Construction Checker, I need to reset the CalledMethods type of the target to top whenever a RMC method is called. This should be nearly a carbon copy of the code in MustCallTransfer I just wrote, but instead of looking up the default type it can just use top. I also need to write tests for this.
- [ ] When re-assigning an owning field, I need to add a check that the containing method is `@ResetMustCall`. I'm not sure if this belongs in the MustCall Checker or the Object Construction Checker. It probably doesn't matter. Write tests for this, too.
- [ ] Write appropriate annotations in the socket stubs, to handle unconnected sockets.
- [ ] Write a test that checks that a target of a RMC method that's a non-owning pointer becomes owning. I think I'll also have to modify the MustCallInvoked checker to get this working.

kelloggm:
Closing in favor of #293

PR #282, opened 2021-01-07
reassignments with the same resource

Nargeshdb:
This PR handles reassignments that the rhs is @MustCallChoice with the lhs.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -512,8 +512,7 @@ class MustCallInvokedChecker {
             // contains mustCallChoiceParamLocal
             ImmutableSet<LocalVarWithTree> setContainingMustCallChoiceParamLocal =
                 getSetContainingAssignmentTreeOfVar(newDefs, mustCallChoiceParamLocal);
-            // If rhs is not MCC with the lhs, then nothing will get filtered by the filter call
-            // because we removed latestAssignmentPair from newDefs earlier
+            // If latestAssignmentPair is null, nothing will get filtered by the filter call
             ImmutableSet<LocalVarWithTree> newSetContainingMustCallChoiceParamLocal =
                 FluentIterable.from(setContainingMustCallChoiceParamLocal)
                     .filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -464,42 +464,32 @@ class MustCallInvokedChecker {
       // Reassignment to the lhs
       if (isVarInDefs(newDefs, (LocalVariableNode) lhs)) {
         ImmutableSet<LocalVarWithTree> setContainingLatestAssignmentPair =
-            getSetContainingAssignmentTreeOfVar(newDefs, (LocalVariableNode) lhs);
+                getSetContainingAssignmentTreeOfVar(newDefs, (LocalVariableNode) lhs);
         LocalVariableNode localPassedAsMCCParam = getLocalPassedAsMustCallChoiceParam(rhs);
         LocalVarWithTree latestAssignmentPair =
-            getAssignmentTreeOfVar(newDefs, (LocalVariableNode) lhs);
-        if (localPassedAsMCCParam != null
-            && setContainingLatestAssignmentPair.stream()
+                getAssignmentTreeOfVar(newDefs, (LocalVariableNode) lhs);
+        if (localPassedAsMCCParam == null
+                || setContainingLatestAssignmentPair.stream()
                 .map(localVarWithTree -> localVarWithTree.localVar.getElement())
-                .anyMatch(elem -> elem.equals(localPassedAsMCCParam.getElement()))) {
-          // If the rhs is MCC with the lhs, replace latestAssignmentPair with this Assignment
-          // Node
-          ImmutableSet<LocalVarWithTree> newSetContainingLatestAssignmentPair =
-              FluentIterable.from(setContainingLatestAssignmentPair)
-                  .filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))
-                  .append(
-                      new LocalVarWithTree(
-                          new LocalVariable((LocalVariableNode) lhs), node.getTree()))
-                  .toSet();
-          newDefs.remove(setContainingLatestAssignmentPair);
-          newDefs.add(newSetContainingLatestAssignmentPair);
-        } else if (setContainingLatestAssignmentPair.size() > 1) {
-          // If the setContainingLatestAssignmentPair has more LocalVarWithTree, remove
-          // latestAssignmentPair
-          ImmutableSet<LocalVarWithTree> newSetContainingLatestAssignmentPair =
-              FluentIterable.from(setContainingLatestAssignmentPair)
-                  .filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))
-                  .toSet();
-          newDefs.remove(setContainingLatestAssignmentPair);
-          newDefs.add(newSetContainingLatestAssignmentPair);
-        } else {
-          // If the setContainingLatestAssignmentPair size is one and the rhs is not MCC with the
-          // lhs
-          checkMustCall(
-              setContainingLatestAssignmentPair,
-              typeFactory.getStoreBefore(node),
-              "variable overwritten by assignment " + node.getTree());
-          newDefs.remove(setContainingLatestAssignmentPair);
+                .noneMatch(elem -> elem.equals(localPassedAsMCCParam.getElement()))) {
+          if (setContainingLatestAssignmentPair.size() > 1) {
+            // If the setContainingLatestAssignmentPair has more LocalVarWithTree, remove
+            // latestAssignmentPair
+            ImmutableSet<LocalVarWithTree> newSetContainingLatestAssignmentPair =
+                    FluentIterable.from(setContainingLatestAssignmentPair)
+                            .filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))
+                            .toSet();
+            newDefs.remove(setContainingLatestAssignmentPair);
+            newDefs.add(newSetContainingLatestAssignmentPair);
+          } else {
+            // If the setContainingLatestAssignmentPair size is one and the rhs is not MCC with the
+            // lhs
+            checkMustCall(
+                    setContainingLatestAssignmentPair,
+                    typeFactory.getStoreBefore(node),
+                    "variable overwritten by assignment " + node.getTree());
+            newDefs.remove(setContainingLatestAssignmentPair);
+          }
         }
       }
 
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -464,21 +465,41 @@ private void handleAssignFromRHS(
...
-        if (setContainingLatestAssignmentPair.size() == 1) {
-          checkMustCall(
-              setContainingLatestAssignmentPair,
-              typeFactory.getStoreBefore(node),
-              "variable overwritten by assignment " + node.getTree());
+        if (localPassedAsMCCParam != null
+            && setContainingLatestAssignmentPair.stream()
+                .map(localVarWithTree -> localVarWithTree.localVar.getElement())
+                .anyMatch(elem -> elem.equals(localPassedAsMCCParam.getElement()))) {
+          // If the rhs is MCC with the lhs, replace latestAssignmentPair with this Assignment
```
msridhar, line 475: I'm not sure I understand what is going on in this case.  Can you write an example here showing why exactly we need this new case?  It seems like code below around line 512 should handle adding the assignment to the set in this case.  Maybe we need to fix a bug down there?

Nargeshdb: There was a bug around line 512! Instead of replacing the previous assignment node, we were just adding the new assignment node.
https://github.com/kelloggm/object-construction-checker/pull/282/commits/4cd278c17fdf8a8d5032bd4b16aeea842d8e3d69

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -490,16 +501,21 @@ private void handleAssignFromRHS(
...
                 getSetContainingAssignmentTreeOfVar(newDefs, mustCallChoiceParamLocal);
-            ImmutableSet<LocalVarWithTree> newImmutableSet =
-                FluentIterable.from(immutableSet)
+            ImmutableSet<LocalVarWithTree> newSetContainingLatestAssignmentPair =
+                FluentIterable.from(setContainingLatestAssignmentPair)
+                    .filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))
                     .append(
                         new LocalVarWithTree(
                             new LocalVariable((LocalVariableNode) lhs), node.getTree()))
                     .toSet();
```
msridhar, line 515: I still can't fully understand this code.  At this point, I think we are handling the case where (1) the `rhs` has a MCC variable `mustCallChoiceParamLocal` that is _already_ being tracked as part of a set, and also (2) `lhs` is also being tracked as part of a set.  (Question: do these have to be the same set?)  In such a case, we want to update the defs to still track a set with both `lhs` and `mustCallChoiceParamLocal`, but the tree for `lhs` should be this latest assignment.  Is this right?

If the above is right, please add a comment to this effect.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -464,21 +465,31 @@ private void handleAssignFromRHS(
...
-              "variable overwritten by assignment " + node.getTree());
-          newDefs.remove(setContainingLatestAssignmentPair);
-        } else {
-          ImmutableSet<LocalVarWithTree> newSetContainingLatestAssignmentPair =
-              FluentIterable.from(setContainingLatestAssignmentPair)
-                  .filter(assignment -> assignment.equals(latestAssignmentPair))
-                  .toSet();
-          newDefs.remove(setContainingLatestAssignmentPair);
-          newDefs.add(newSetContainingLatestAssignmentPair);
+        if (localPassedAsMCCParam == null
```
msridhar, line 471: add a comment here saying the MCC case is handled below?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -490,16 +501,21 @@ private void handleAssignFromRHS(
...
                         new LocalVarWithTree(
                             new LocalVariable((LocalVariableNode) lhs), node.getTree()))
                     .toSet();
-            newDefs.remove(immutableSet);
-            newDefs.add(newImmutableSet);
+
+            newDefs.remove(setContainingLatestAssignmentPair);
+            newDefs.add(newSetContainingLatestAssignmentPair);
           }
         } else {
```
msridhar, line 520: Let's add a check under the `else` that `lhs` is not already part of `newDefs`, and if it is, throw a `RuntimeException` (to ensure our logic is correct)

msridhar (changes requested):
Looking good, just needs a couple of comments

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: > I still can't fully understand this code. At this point, I think we are handling the case where (1) the `rhs` has a MCC variable `mustCallChoiceParamLocal` that is _already_ being tracked as part of a set, and also (2) `lhs` is also being tracked as part of a set. (Question: do these have to be the same set?) 
>
Not necessarily. If these are in the same set, we will replace the previous assignment tree with the current assignment tree. Otherwise, `filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))` does nothing and then the currect assignment tree will be added to the set that contains `mustCallChoiceParamLocal`.
>In such a case, we want to update the defs to still track a set with both `lhs` and `mustCallChoiceParamLocal`, but the tree for `lhs` should be this latest assignment. Is this right?
>
It's true when these are in the same set.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: We shouldn't throw an Exception here! If it's not part of `newDefs`, it means we shouldn't track this resource (e.g. this resource is MCC with `@NotOwning` parameters)!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/282/commits/2dc62e3a4d37190f4881de233fdaa7fe334c8e53

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/282/commits/2dc62e3a4d37190f4881de233fdaa7fe334c8e53

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: Sorry, I meant we should throw an exception if `lhs` _is_ already in `newDefs`.  I think that should be impossible under this `else` case, which is why the code just adds the new `LocalVarWithTree` for `lhs` without checking if it's already there.  Or did I miss something?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -490,16 +504,24 @@ private void handleAssignFromRHS(
...
+            // lhs), then we will replace the previous assignment tree of lhs with current
+            // assignment tree, otherwise we will just add a new LocalVarWithTree to the set that
+            // contains mustCallChoiceParamLocal
+            ImmutableSet<LocalVarWithTree> setContainingMustCallChoiceParamLocal =
                 getSetContainingAssignmentTreeOfVar(newDefs, mustCallChoiceParamLocal);
-            ImmutableSet<LocalVarWithTree> newImmutableSet =
-                FluentIterable.from(immutableSet)
+            ImmutableSet<LocalVarWithTree> newSetContainingMustCallChoiceParamLocal =
+                FluentIterable.from(setContainingMustCallChoiceParamLocal)
+                    .filter(Predicates.not(Predicates.equalTo(latestAssignmentPair)))
```
msridhar, line 517: Is the idea here that if `latestAssignmentPair` is `null`, nothing will get filtered by the `filter` call, which is the right thing to do?  If so, we should document that.  Right now the code is a bit too clever 🙂

msridhar (commented):
A couple more questions

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: Sorry, I didn't read it carefully. 
https://github.com/kelloggm/object-construction-checker/pull/282/commits/171817635c00a3c0b87306ba2feb7fae90f9bb92

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -490,18 +504,29 @@ private void handleAssignFromRHS(
...
                     .toSet();
-            newDefs.remove(immutableSet);
-            newDefs.add(newImmutableSet);
+
+            newDefs.remove(setContainingMustCallChoiceParamLocal);
+            newDefs.add(newSetContainingMustCallChoiceParamLocal);
           }
         } else {
+          if (isVarInDefs(newDefs, (LocalVariableNode) lhs)) {
+            throw new RuntimeException();
```
msridhar, line 528: Add a message to the exception, "did not expect lhs to be in newDefs"

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: @Nargeshdb any thoughts on this one?

msridhar (changes requested):

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: > Is the idea here that if `latestAssignmentPair` is `null`, nothing will get filtered by the `filter` call, which is the right thing to do?
>
Yes, exactly.
> If so, we should document that. Right now the code is a bit too clever 🙂
>
I added a comment here.

msridhar: Sorry I meant a comment specifically about the null case

Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/282/commits/b3b069b47173c8ec3a3cb8e34cd988d7be04a85a

msridhar (approved):
👍

PR #291, opened 2021-01-22
Introduce temporary variables for calls and ternary expressions

Nargeshdb:


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -31,6 +35,7 @@
  */
 public class ObjectConstructionAnnotatedTypeFactory extends CalledMethodsAnnotatedTypeFactory {
 
+  protected BiMap<LocalVariableNode, Node> biMap = HashBiMap.create();
```
msridhar, line 38: This field needs a better name than `biMap` and some documentation

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -73,6 +78,9 @@ public void postAnalyze(ControlFlowGraph cfg) {
   List<String> getMustCallValue(Tree tree) {
     MustCallAnnotatedTypeFactory mustCallAnnotatedTypeFactory =
         getTypeFactoryOfSubchecker(MustCallChecker.class);
+    if (mustCallAnnotatedTypeFactory == null) {
```
msridhar, line 81: How is this change related to temporary variables?  Same for the next change below

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -45,6 +55,8 @@
 public class ObjectConstructionTransfer extends CalledMethodsTransfer {
   private final ObjectConstructionAnnotatedTypeFactory atypefactory;
 
+  public final TreeBuilder treeBuilder;
```
msridhar, line 58: Should probably be `private`?  Also document what this is for

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -74,9 +88,54 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
             result.getElseStore(),
             exceptionalStores);
     exceptionalStores = null;
+
+    Node receiver = node.getTarget().getReceiver();
+    addTemporaryVars(result, node);
+    if (atypefactory.biMap.inverse().containsKey(receiver)) {
```
msridhar, line 94: To me this code would be clearer if you stored the result of `atypefactory.biMap.inverse().get(receiver)` in a local variable, and then changed the condition to check if that variable is null.  Then you could avoid writing `atypefactory.biMap.inverse()` in two places

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -74,9 +88,54 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
...
 
+  @Override
+  public TransferResult<CFValue, CFStore> visitObjectCreation(
+      ObjectCreationNode node, TransferInput<CFValue, CFStore> input) {
+    TransferResult<CFValue, CFStore> result = super.visitObjectCreation(node, input);
+    addTemporaryVars(result, node);
+    return result;
+  }
+
+  private void addTemporaryVars(TransferResult<CFValue, CFStore> result, Node node) {
```
msridhar, line 113: Please add Javadoc

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -74,9 +88,54 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
...
+    if (atypefactory.hasMustCall(node.getTree())) {
+      if (atypefactory.hasMustCall(node.getTree())) {
```
msridhar, line 115: Redundant with previous line

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
msridhar: Also this method needs a better name.  Maybe `updateStoreWithTempVar`?  I think the method either creates or looks up the temp var `t` for `node`, and then updates the store to give `t` the same type as `node`.  Is that correct?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -74,9 +88,54 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
...
+        LocalVariableNode localVariableNode;
+        if (!atypefactory.biMap.inverse().containsKey(node)) {
+          VariableTree temp = createTemporaryVar(node);
+          IdentifierTree identifierTree = treeBuilder.buildVariableUse(temp);
+          localVariableNode = new LocalVariableNode(identifierTree);
+          localVariableNode.setInSource(true);
+          atypefactory.biMap.put(localVariableNode, node);
+        } else {
+          localVariableNode = atypefactory.biMap.inverse().get(node);
+        }
```
msridhar, line 126: I would extract this code into its own method, `getOrCreateTempVar`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -74,9 +88,54 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
             result.getElseStore(),
             exceptionalStores);
     exceptionalStores = null;
+
+    Node receiver = node.getTarget().getReceiver();
+    addTemporaryVars(result, node);
```
msridhar, line 93: I'd move this line before the previous one, to keep the code about the `receiver` together.  Then, add some comments indicating what the code involving `receiver` is doing

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -167,9 +227,41 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
         }
       }
+      AnnotationMirror newAnno = atypefactory.createAccumulatorAnnotation(valuesAsList);
+      exceptionalStores.values().stream().forEach(s -> s.insertValue(target, newAnno));
     }
-    AnnotationMirror newAnno = atypefactory.createAccumulatorAnnotation(valuesAsList);
-    exceptionalStores.values().stream().forEach(s -> s.insertValue(target, newAnno));
+  }
+
+  private void insertIntoStores(
```
msridhar, line 235: Is this method copy-pasted from somewhere else?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -167,9 +227,41 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
+            uniqueName("tmp"),
+            ee.getEnclosingElement(),
+            (ExpressionTree) method.getTree());
+    return iteratorVariable;
+  }
+
+  protected long uid = 0;
+
+  protected String uniqueName(String prefix) {
+    return prefix + "#num" + uid++;
```
msridhar, line 264: This code seems to be copy-pasted based on local variable names.  It may be hard to share the logic, but at least rename the variables to be appropriate to this usage.

object-construction-checker/tests/mustcall/ACMethodInvocationTest.java
```diff
@@ -32,17 +32,11 @@ void c() {}
         return f;
     }
 
-    //TODO
     void CallMethodsInSequence() {
-        // won't pass yet due to dataflow issue
-        // :: error: required.method.not.called
```
msridhar, line 38: 🎉 for both these tests passing!

object-construction-checker/tests/socket/ACSocketTest.java
```diff
@@ -214,19 +214,6 @@ void replaceVarWithNull(String address, int port) {
 
     }
 
-//    @EnsuresCalledMethodsIf(expression = "#1", methods = {"close"}, result = true)
```
msridhar, line 217: why is this deleted?

object-construction-checker/tests/socket/ACSocketTest.java
```diff
@@ -333,72 +320,6 @@ void makePrependableSocket() throws IOException {
         final PrependableSocket prependableSocket = new PrependableSocket(null);
     }
 
-
-
-
-
-//    private void acceptConnections() {
```
msridhar, line 340: why is all this commented out?

unconnected-socket-checker/src/main/java/org/checkerframework/checker/unconnectedsocket/UnconnectedSocketTransfer.java
```diff
@@ -7,8 +7,9 @@
 import org.checkerframework.dataflow.cfg.node.FieldAccessNode;
 import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
 import org.checkerframework.dataflow.cfg.node.Node;
-import org.checkerframework.dataflow.expression.FlowExpressions;
-import org.checkerframework.dataflow.expression.Receiver;
+// import org.checkerframework.dataflow.expression.FlowExpressions;
```
msridhar, line 10: we should get rid of these commented imports

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -219,16 +219,50 @@ private void checkPseudoAssignToOwning(Node node) {
...
-      AnnotationMirror cmAnno =
-          typeFactory.getAnnotatedType(tree).getAnnotationInHierarchy(typeFactory.top);
-      if (!calledMethodsSatisfyMustCall(mustCallVal, cmAnno)) {
-        checker.reportError(
-            tree,
-            "required.method.not.called",
-            MustCallInvokedChecker.formatMissingMustCallMethods(mustCallVal),
-            TreeUtils.typeOf(tree).toString(),
-            "never assigned to an @Owning location");
+      if (typeFactory.biMap.inverse().containsKey(node)) {
```
msridhar, line 222: This code definitely needs some comments; I can't immediately see what it is doing.  Also, I was imagining that this whole `checkPseudoAssignToOwning` method would go away once temporary variables are properly implemented.  Do you agree?

msridhar (commented):
Some initial comments, mostly around documentation and code cleanness.  Unfortunately CI won't pass until the relevant changes get merged into a Checker Framework release; please confirm that `./gradlew build` passes locally

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
Nargeshdb: That's correct, I updated the name and added Javadoc:
https://github.com/kelloggm/object-construction-checker/pull/291/files#diff-3fc5dcd77eafbbf61a8da8156fac14c917b05d258d829aab8e8706b815b91b52R137

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/291/files#diff-3fc5dcd77eafbbf61a8da8156fac14c917b05d258d829aab8e8706b815b91b52R148

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
Nargeshdb: Yes, It's from AccumulationTransfer class.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
Nargeshdb: I renamed the variables to "temp-var-id"

build.gradle
```diff
@@ -12,7 +12,7 @@ plugins {
 
 
 ext.versions = [
-    checkerFramework: "3.8.0",
+    checkerFramework: "3.9.2-SNAPSHOT",
```
msridhar, line 15: Make this `3.10.0` I think

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -31,6 +34,11 @@
  */
 public class ObjectConstructionAnnotatedTypeFactory extends CalledMethodsAnnotatedTypeFactory {
 
+  /**
+   * Bidirectional map to preserve temporal variables created for nodes with non-empty @MustCall
+   * annotation and the corresponding nodes
+   */
+  protected BiMap<LocalVariableNode, Tree> mapTempVarToNode = HashBiMap.create();
```
msridhar, line 41: Just `tempVarToNode`:
```suggestion
  protected BiMap<LocalVariableNode, Tree> tempVarToNode = HashBiMap.create();
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
msridhar: I still have this question

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
```diff
@@ -87,6 +99,9 @@ public void postAnalyze(ControlFlowGraph cfg) {
         getTypeFactoryOfSubchecker(MustCallChecker.class);
     AnnotatedTypeMirror mustCallAnnotatedType =
         mustCallAnnotatedTypeFactory.getAnnotatedType(element);
+    if (mustCallAnnotatedTypeFactory == null) {
```
msridhar, line 102: How is this change related to temporary variables?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -45,6 +57,9 @@
 public class ObjectConstructionTransfer extends CalledMethodsTransfer {
   private final ObjectConstructionAnnotatedTypeFactory atypefactory;
 
+  /** TreeBuilder for building new AST nodes */
```
msridhar, line 60:
```suggestion
  /** TreeBuilder for building new AST nodes, for creation of temporary variables */
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -58,6 +73,16 @@
...
     super(analysis);
     this.atypefactory = (ObjectConstructionAnnotatedTypeFactory) analysis.getTypeFactory();
+    ProcessingEnvironment env = atypeFactory.getChecker().getProcessingEnvironment();
+    treeBuilder = new TreeBuilder(env);
+  }
+
+  @Override
+  public TransferResult<CFValue, CFStore> visitTernaryExpression(
+      TernaryExpressionNode node, TransferInput<CFValue, CFStore> input) {
+    TransferResult<CFValue, CFStore> result = super.visitTernaryExpression(node, input);
```
msridhar, line 83: Not exactly related to this code, but did we add a test for calling `close()` on a ternary expression directly?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -74,9 +99,65 @@ public ObjectConstructionTransfer(final CFAnalysis analysis) {
             result.getElseStore(),
             exceptionalStores);
     exceptionalStores = null;
+
+    updateStoreWithTempVar(result, node);
+
+    // If the receiver exists in the mapTempVarToNode, then its temporal variable's type will be
+    // updated.
```
msridhar, line 106:
```suggestion
    // If there is a temporary variable for the receiver, update its type.
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
msridhar: can we find a way to not do copy-paste?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -167,9 +249,49 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
+      CFStore store = result.getRegularStore();
+      store.insertValue(target, newAnno);
     }
-    AnnotationMirror newAnno = atypefactory.createAccumulatorAnnotation(valuesAsList);
-    exceptionalStores.values().stream().forEach(s -> s.insertValue(target, newAnno));
+  }
+
+  protected VariableTree createTemporaryVar(Node method) {
+    ExpressionTree tree = (ExpressionTree) method.getTree();
+    TypeMirror iteratorType = TreeUtils.typeOf(tree);
```
msridhar, line 272:
```suggestion
    TypeMirror treeType = TreeUtils.typeOf(tree);
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -167,9 +249,49 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
+    Element enclosingElement;
+    TreePath path = atypefactory.getPath(tree);
+    if (path == null) {
+      enclosingElement = TreeUtils.elementFromTree(tree).getEnclosingElement();
+    } else {
+      Tree methodTree = TreePathUtil.enclosingMethod(path);
+      enclosingElement = TreeUtils.elementFromTree(methodTree);
+    }
+    // Declare and initialize a new, unique iterator variable
+    VariableTree iteratorVariable =
```
msridhar, line 282:
```suggestion
    VariableTree tmpVarTree =
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -690,6 +586,26 @@ private void handleSuccessorBlocks(
             setAssign.stream()
                 .allMatch(assign -> succRegularStore.getValue(assign.localVar) == null);
         if (succ instanceof SpecialBlockImpl || noSuccInfo) {
+
+          // Remove the temporary variable defined for a node that throws an exception from the
+          // exceptional successors
+          if (succAndExcType.second != null) {
+            Node exceptionalNode = removeCasts(((ExceptionBlockImpl) block).getNode());
```
msridhar, line 593:
```suggestion
            Node exceptionalNode = removeCasts(((ExceptionBlock) block).getNode());
```

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -123,129 +119,89 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
           handleReturn((ReturnNode) node, cfg, newDefs);
         } else if (node instanceof MethodInvocationNode || node instanceof ObjectCreationNode) {
           handleInvocation(newDefs, node);
-        } else if (node instanceof TypeCastNode || node instanceof TernaryExpressionNode) {
```
msridhar, line 126: Why don't we need to do anything for `TypeCastNode` anymore?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -123,129 +119,89 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
-    }
-    if (!nestedInCastOrTernary(node) && hasNestedPseudoAssignInvoke(node, defs)) {
-      checkPseudoAssignToOwning(node);
+  private void handleTernary(Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
+    LocalVariableNode ternaryLocal = typeFactory.mapTempVarToNode.inverse().get(node.getTree());
+
+    Node operand = removeCasts(((TernaryExpressionNode) node).getThenOperand());
+    LocalVariableNode operandLocal = typeFactory.mapTempVarToNode.inverse().get(operand.getTree());
+    if (operandLocal == null || !isVarInDefs(defs, operandLocal)) {
+      operand = removeCasts(((TernaryExpressionNode) node).getElseOperand());
```
msridhar, line 137: It seems like we never look at the else operand if we have a  local for the then operand.  That's not what I expected; I thought we'd handle them separately, treating each as a separate assignment to the temporary for the ternary expression.  Did I misunderstand?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -123,129 +119,89 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
+    updateDefsWithTempVar(defs, node);
   }
 
-  /**
-   * Given a node representing a method or constructor call, checks that if the call has a non-empty
-   * {@code @MustCall} type, then its result is pseudo-assigned to some location that can take
-   * ownership of the result
-   */
-  private void checkPseudoAssignToOwning(Node node) {
+  private void updateDefsWithTempVar(Set<ImmutableSet<LocalVarWithTree>> defs, Node node) {
```
msridhar, line 165: This method needs comments.  I can't follow the logic.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -255,32 +211,17 @@ private boolean isOwningAssignmentLhs(Element elem) {
    * a super constructor call, the method's return type is annotated {@link NotOwning}, or the
    * receiver parameter is in the {@code defs} and has @MustCallChoice annotation.
    */
-  private boolean shouldSkipInvokePseudoAssignCheck(
-      Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
+  private boolean shouldSkipInvokePseudoAssignCheck(Node node) {
```
msridhar, line 214: rename this method?  since we're not doing pseudoassign checks anymore?

object-construction-checker/tests/socket/ACSocketTest.java
msridhar: still have the same question

object-construction-checker/tests/socket/GetChannelOnLocks.java
```diff
@@ -0,0 +1,36 @@
+import java.io.IOException;
+import java.nio.channels.FileLock;
+
+class GetChannelOnLocks {
```
msridhar, line 4: Add comments on what we are testing for here

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -123,129 +119,89 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
-                        NotOwning.class)
-                    == null
-                : typeFactory.getDeclAnnotation(elementForType, Owning.class) != null;
-      } else if (assignmentContext instanceof LambdaReturnContext) {
-        // TODO handle this case.  For now we will report an error
-      } else {
-        throw new BugInCF("unexpected AssignmentContext type " + assignmentContext.getClass());
+    if (typeFactory.mapTempVarToNode.inverse().containsKey(node.getTree())) {
+
+      LocalVariableNode temporaryLocal = typeFactory.mapTempVarToNode.inverse().get(node.getTree());
```
msridhar, line 169: Move this line before the `if` check, and change the `if` check to see if `temporaryLocal == null`

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -123,129 +119,89 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
-            MustCallInvokedChecker.formatMissingMustCallMethods(mustCallVal),
-            TreeUtils.typeOf(tree).toString(),
-            "never assigned to an @Owning location");
+
+      if (receiver == null
+          && node instanceof MethodInvocationNode
+          && (typeFactory.returnsThis((MethodInvocationTree) tree))) {
+        receiver = ((MethodInvocationNode) node).getTarget().getReceiver();
+        if (receiver instanceof MethodInvocationNode) {
+          receiver = typeFactory.mapTempVarToNode.inverse().get(receiver.getTree());
```
msridhar, line 183: We should add a method `getTempVarForTree()` to the type factory that encapsulates this logic of doing a `get()` on the `inverse()`, and use that everywhere

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -123,129 +119,89 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
...
-        // TODO handle this case.  For now we will report an error
-      } else {
-        throw new BugInCF("unexpected AssignmentContext type " + assignmentContext.getClass());
+    if (typeFactory.mapTempVarToNode.inverse().containsKey(node.getTree())) {
+
+      LocalVariableNode temporaryLocal = typeFactory.mapTempVarToNode.inverse().get(node.getTree());
+      LocalVarWithTree lhsLocalVarWithTreeNew =
+          new LocalVarWithTree(new LocalVariable(temporaryLocal), tree);
+
+      Node receiver = null;
```
msridhar, line 173: It's here that I start to get lost on what the method is doing

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -445,107 +343,100 @@ private Node removeCasts(Node node) {
   }
 
   private void handleAssignFromRHS(
-      AssignmentNode node, Set<ImmutableSet<LocalVarWithTree>> newDefs, Node rhs) {
+      AssignmentNode node, Set<ImmutableSet<LocalVarWithTree>> newDefs, Node rhstemp) {
     Node lhs = node.getTarget();
     Element lhsElement = TreeUtils.elementFromTree(lhs.getTree());
+    if (typeFactory.mapTempVarToNode.inverse().containsKey(rhstemp.getTree())) {
```
msridhar, line 349: This logic is weird.  Why did we rename the parameter to `rhstemp`, which is kind of strange?  Can we write this code so it doesn't re-assign the parameter?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -724,6 +640,10 @@ private void handleSuccessorBlocks(
     }
   }
 
+  private boolean isTempVar(LocalVariable localVariable) {
+    return localVariable.getElement().getSimpleName().toString().contains("temp-var");
```
msridhar, line 644: can't we add a method to the type factory to check this properly by checking for membership in the bi-map?

msridhar (changes requested):
Looking promising!  I have a few comments

msridhar:
Another thought.  Would this code become even simpler if we created temporary variables for downcasts as well?  There are various places where we unwrap downcasts that I think could be simplified.

build.gradle
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/291/commits/790ac07e99314570d32d4c8b2fc03158226ec15a

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/291/commits/790ac07e99314570d32d4c8b2fc03158226ec15a

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: I added this check to handle `NullPointerException` [here](https://github.com/kelloggm/object-construction-checker/blob/temporaryVar/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java#L103) when the given `node` is `(this).<init>()`. I got a lot of `NullPointerException` in AutoValue test cases.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
Nargeshdb: Can we make this method public in AccumulationTransfer?

object-construction-checker/tests/socket/ACSocketTest.java
Nargeshdb: I thought it's repetitive but it's not, so I added it again.
https://github.com/kelloggm/object-construction-checker/pull/291/commits/32c6c64d3fe2e77413bc1fadc0878fd07aea1cfd

object-construction-checker/tests/socket/ACSocketTest.java
Nargeshdb: It was commented out, I just removed it because it contains some methods/fields that are not defined in this class, so we can't test it. I added it again https://github.com/kelloggm/object-construction-checker/pull/291/commits/03e52aad72ffba8808e8d6ef71189925d887c288

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/291/files/424f947ae09c7edf540ad6c5c9ad821549ab31ed#r570794811

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
Nargeshdb: I removed this method because I don't use it.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/291/files#diff-d9c49afa35f8d4df342e326e80fd73fd3d5df0b9c0d3a1c95cd6662f84251d00R356

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
msridhar: @Nargeshdb we can make it `protected` and then call it from here

msridhar (changes requested):
Just one minor thing.  We need to remember as follow-ups to (1) fix the known false positives from weird ternary expressions and (2) also introduce temporary variables for downcasts

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
msridhar: This can be a follow up change as it requires a change in CF

PR #293, opened 2021-01-30
ResetMustCall support

kelloggm:
This supersedes #280 (i.e. it includes the same functionality). It also replaces the Unconnected Socket Checker with a sound approach based on `@ResetMustCall`.

For now, the tests won't pass, because this relies on https://github.com/typetools/checker-framework/pull/4189. I doubt that will make it into the Feb 1 release (since it's been waiting for review from @mernst for days with no response, and he hasn't been responding to my messages), so we'll need to use local CF versions and test carefully rather than relying on CI for the next month (once this is merged). Sorry :(


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -184,6 +206,11 @@ private void handleInvocation(Set<ImmutableSet<LocalVarWithTree>> newDefs, Node
    */
   private void checkPseudoAssignToOwning(Node node) {
     Tree tree = node.getTree();
+    // This call gets the declared must-call type of the return value, which isn't affected by
+    // dataflow.
+    // Therefore, it's safe to just call the version of getMustCallValue that indirects to
+    // getAnnotatedType,
+    // and not the version that uses the mustcall store directly.
```
msridhar, line 213: Nit: newlines are placed kind of oddly in this block comment

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -258,6 +285,11 @@ private boolean isOwningAssignmentLhs(Element elem) {
   private boolean shouldSkipInvokePseudoAssignCheck(
       Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
     Tree callTree = node.getTree();
+    // This call gets the declared must-call type of the return value, which isn't affected by
+    // dataflow.
+    // Therefore, it's safe to just call the version of getMustCallValue that indirects to
+    // getAnnotatedType,
+    // and not the version that uses the mustcall store directly.
```
msridhar, line 292: Nit: newlines are placed kind of oddly in this block comment

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -543,12 +585,148 @@ private void handleAssignFromRHS(
...
+   */
+  private void checkEnclosingMethodIsResetMC(AssignmentNode node) {
+    Node lhs = node.getTarget();
+    if (!(lhs instanceof FieldAccessNode)) {
+      return;
+    }
+
+    String receiverString = receiverAsString((FieldAccessNode) lhs);
+
+    TreePath currentPath = typeFactory.getPath(node.getTree());
```
msridhar, line 679: Is there any way to get the top-level tree for a CFG given some node in that CFG, which should be the desired `MethodTree`?  Just curious; I'm getting that getting the path for a tree might be expensive (requiring a search of the AST for the whole compilation unit)

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -543,12 +585,148 @@ private void handleAssignFromRHS(
...
+  private String receiverAsString(FieldAccessNode lhs) {
+    Node receiver = lhs.getReceiver();
+    if (receiver instanceof ThisNode) {
+      return "this";
+    }
+    if (receiver instanceof LocalVariableNode) {
+
+      return ((LocalVariableNode) receiver).getName();
+    }
+    throw new BugInCF(
```
msridhar, line 726: Should we really be crashing in this case?  Can't we just report an `incompatible.reset.mustcall` error?

msridhar (changes requested):
Overall this seems great!  Just a few questions

build.gradle
```diff
@@ -21,8 +21,8 @@ subprojects { project ->
     repositories {
         // Use jcenter for resolving your dependencies.
         // You can declare any Maven/Ivy/file repository here.
-        jcenter()
         mavenLocal()
+        jcenter()
```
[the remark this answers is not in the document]
kelloggm: No. I'm not sure why I did this - probably just a mistake. I'll undo it.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallChecker.java
```diff
@@ -23,6 +23,10 @@ public Properties getMessagesProperties() {
     messages.setProperty(
         "inconsistent.mustcall.subtype",
         "%s is annotated as %s, but one of its supertypes has an %s annotation, which is inconsistent.");
+    messages.setProperty(
+        "mustcall.not.parseable",
+        "The method %s is annotated as @ResetMustCall, but the target (%s) was unparseable in the current context. "
```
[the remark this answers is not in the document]
kelloggm: I agree, but couldn't come up with a bette option. The second part of the message ("extract the unparseable expression into a local variable") is at least actionable.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: There might be, but the path is needed for viewpoint adaptation later in the method regardless so computing it is unavoidable.

It might be possible to defer computing the path until after we check for the presence of an `@ResetMustCall` annotation on the declaration, if we were able to get the `MethodTree` another way - which would probably make it worthwhile.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: I originally wrote this as a crash to try to figure out what else it might be, but it was never anything else besides "this" or a LocalVariable in our tests.

I think I'd prefer to leave it as a crash, since if we ever encounter that crash we'll be able to modify this method to handle that kind of receiver - that's better than silently failing, IMO.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: Ok.  We don't need to block landing on this, but maybe we should add a comment if we think there may be a better way to do it in the future.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: Sure, that sounds good for now

msridhar (commented):
I think in terms of the changes here I am ok with them.  But before landing, we should discuss how we will stage this change alongside #291, and also whether we want to wait until the relevant Checker Framework changes land before landing this.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: I'll add a comment.

kelloggm:
@msridhar while writing up the design of this feature for the paper, I realized that I forgot to implement one of the cases I described in my original design email: calls to ResetMustCall methods on non-owning pointers must be forbidden! Otherwise, such a call would create an untracked obligation.

I've modified the PR to cover this case, with my latest commit. I'd appreciate it if you could take a look at the new code when you get a chance.

Thanks!

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -199,6 +202,94 @@ private void handleInvocation(Set<ImmutableSet<LocalVarWithTree>> newDefs, Node
...
+      Element elt = ((FieldAccess) target).getField();
+      if (typeFactory.getDeclAnnotation(elt, Owning.class) != null) {
+        // if the target is an Owning field, this satisfies case 1
+        return;
+      }
+    }
+
+    MethodTree containingMethod = TreePathUtil.enclosingMethod(currentPath);
+    if (containingMethod != null) {
+      ExecutableElement enclosingMethod = TreeUtils.elementFromDeclaration(containingMethod);
```
msridhar, line 271: Variable names `containingMethod` and `enclosingMethod` are a bit confusing.  Maybe `enclosingMethod` and `enclosingMethodElement`?

msridhar (approved):
new changes look good, just one nit

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: Thanks for the suggestion. I also made a similar change in another part of this PR that does something similar and used the same confusing naming scheme.

PR #301, opened 2021-02-09
Fix crash in Zookeeper

kelloggm:
Fix is easy, and the crash would be hard to repro, so this is just the fix. I can try to minimize the failure if we think it's critical to have a unit test. Stack trace started with:

```
  Exception: java.util.NoSuchElementException: No value present; java.util.NoSuchElementExc\
eption: No value present
        at java.util.Optional.get(Optional.java:135)
        at org.checkerframework.checker.objectconstruction.ObjectConstructionAnnotatedTypeF\
actory.getMustCallValue(ObjectConstructionAnnotatedTypeFactory.java:113)
        at org.checkerframework.checker.objectconstruction.MustCallInvokedChecker.checkMust\
Call(MustCallInvokedChecker.java:997)
        at org.checkerframework.checker.objectconstruction.MustCallInvokedChecker.handleSuc\
cessorBlocks(MustCallInvokedChecker.java:898)
        at org.checkerframework.checker.objectconstruction.MustCallInvokedChecker.checkMust\
CallInvoked(MustCallInvokedChecker.java:143)
        at org.checkerframework.checker.objectconstruction.ObjectConstructionAnnotatedTypeF\
actory.postAnalyze(ObjectConstructionAnnotatedTypeFactory.java:79)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericA\
nnotatedTypeFactory.java:1506)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.performFlowAnaly\
sis(GenericAnnotatedTypeFactory.java:1352)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.checkAndPerformF\
lowAnalysis(GenericAnnotatedTypeFactory.java:1783)
        at org.checkerframework.framework.type.GenericAnnotatedTypeFactory.preProcessClassT\
ree(GenericAnnotatedTypeFactory.java:386)
```

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -146,7 +146,12 @@ class MustCallInvokedChecker {
 
   private void handleTernary(Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
     LocalVariableNode ternaryLocal = typeFactory.getTempVarForTree(node);
-
+    // This can happen when one side of the ternary is null, I think.
+    // Without this, we get a crash in tests/mustcall/ZookeeperTernaryCrash.java
+    // on the only ternary in that program snippet.
+    if (ternaryLocal == null) {
+      return;
+    }
     // First check then operand
     Node operand = removeCasts(((TernaryExpressionNode) node).getThenOperand());
     LocalVariableNode operandLocal = typeFactory.getTempVarForTree(operand);
@@ -996,7 +1001,7 @@ class MustCallInvokedChecker {
 
     List<String> mustCallValue = typeFactory.getMustCallValue(localVarWithTreeSet, mcStore);
     // optimization: if there are no must-call methods, we do not need to perform the check
-    if (mustCallValue.isEmpty()) {
+    if (mustCallValue == null || mustCallValue.isEmpty()) {
       return;
     }
 
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionAnnotatedTypeFactory.java
@@ -92,7 +92,7 @@ public class ObjectConstructionAnnotatedTypeFactory extends CalledMethodsAnnotat
    *     then the default MustCall type of each variable's class will be used.
    * @return the list of must-call method names
    */
-  public List<String> getMustCallValue(
+  public @Nullable List<String> getMustCallValue(
       ImmutableSet<LocalVarWithTree> localVarWithTreeSet, @Nullable CFStore mcStore) {
     MustCallAnnotatedTypeFactory mustCallAnnotatedTypeFactory =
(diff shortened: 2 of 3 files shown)


kelloggm:
@msridhar the original fix was insufficient to actually prevent a crash; it just changed the cause. I factored out a test case and included it here, along with the several fixes in our null-handling that lets the test pass.

msridhar (commented):
The change looks ok in terms of preventing crashes, but I am concerned it may lead to false negatives.  @Nargeshdb can you take a look?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -146,7 +146,12 @@ void checkMustCallInvoked(ControlFlowGraph cfg) {
 
   private void handleTernary(Node node, Set<ImmutableSet<LocalVarWithTree>> defs) {
     LocalVariableNode ternaryLocal = typeFactory.getTempVarForTree(node);
-
+    // This can happen when one side of the ternary is null, I think.
+    // Without this, we get a crash in tests/mustcall/ZookeeperTernaryCrash.java
+    // on the only ternary in that program snippet.
+    if (ternaryLocal == null) {
+      return;
```
Nargeshdb, line 153: I think we are creating a temporary variable for all ternary expressions in ObjectConstructionTransfer.java, so why is it null in this case?

kelloggm: I don't know. I suspect it might have something to do with wildcards?

kelloggm: or possibly that the other side of that ternary is `null`?

Nargeshdb: But we could handle [this](https://github.com/kelloggm/object-construction-checker/blob/e51bc0304283c5734be161df7b096c06df66c9b6/object-construction-checker/tests/mustcall/ACRegularExitPointTest.java#L349) test! In this test case null in then operand.

Nargeshdb: We don't add a temporary varibale for primitive types into the store ([here](https://github.com/kelloggm/object-construction-checker/blob/orElseNull/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java#L159)) and that's why sometimes `ternaryLocal` is null. In this example:
`final Integer type = entry.size() >= 2 ? (Integer) entry.get(0) : null;`
ternary expression has primitive type in the then branch and we don't have a temporary vairbale for it in the store.

Nargeshdb (approved):

Nargeshdb:
> The change looks ok in terms of preventing crashes, but I am concerned it may lead to false negatives. @Nargeshdb can you take a look?

These crashes are because of primitive types and can't lead to false negatives.

Nargeshdb:
> > The change looks ok in terms of preventing crashes, but I am concerned it may lead to false negatives. @Nargeshdb can you take a look?
> 
> These crashes are because of primitive types and can't lead to false negatives.

I've tested this branch on hadoop and we didn't have any crashes.

PR #304, opened 2021-02-10
Counting must-call obligations

Nargeshdb:
This PR handles the negative count issue.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -185,6 +185,7 @@ class MustCallInvokedChecker {
                 ResetMustCall.class)
             != null) {
       checkResetMustCallInvocation(defs, (MethodInvocationNode) node);
+      incrementNumMustCall(node.getTree());
     }
 
     if (shouldSkipInvokeCheck(node)) {
@@ -192,7 +193,7 @@ class MustCallInvokedChecker {
     }
 
     if (typeFactory.hasMustCall(node.getTree())) {
-      incrementNumMustCall();
+      incrementNumMustCall(node.getTree());
     }
     updateDefsWithTempVar(defs, node);
   }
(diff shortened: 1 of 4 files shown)


msridhar (commented):
Since this involves `@ResetMustCall` I'd like @kelloggm to take a look

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -185,12 +185,15 @@ private void handleInvocation(Set<ImmutableSet<LocalVarWithTree>> defs, Node nod
                 ResetMustCall.class)
             != null) {
       checkResetMustCallInvocation(defs, (MethodInvocationNode) node);
-      incrementNumMustCall();
```
kelloggm, line 188: This doesn't match my intuition - I would think that calling a reset-must-call method should create a new obligation. But we could also view it as part of the old obligation - this is really a question of how we'd like it to count.

I'm fine with either way, but I'd prefer we increment the count here.

msridhar:
Chatted with @Nargeshdb about this.  We outlined the following principles:

* We should collect a set _MC_ of `Node`s (or `Tree`s, equivalently?) that introduce a must-call obligation.  The final must-call count should equal the size of this set.  (We want a set to avoid double-counting, since the worklist alg can visit a `Node` more than once.)
* An object creation node or method invocation node should be _MC_ if its return type has a non-empty `@MustCall` qualifier.
* A method invocation node should be included if it has a `@ResetMustCall` annotation.
* Should an invocation be included if it is `@MustCallChoice` and causes a new variable to be tracked?  Not sure on this one, I am ok either way.

@Nargeshdb is working on updating the PR, and I think she will increment the count for `@MustCallChoice` for now.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -35,10 +38,10 @@
   public static final String COUNT_MUST_CALL = "countMustCall";
 
   /**
-   * The number of expressions with must-call obligations that were checked. Incremented only if the
+   * Set containing expressions with must-call obligations that were checked. Increment only if the
    * {@link #COUNT_MUST_CALL} option was supplied.
    */
-  int numMustCall = 0;
+  Set<Tree> mustCallObligations = new HashSet<>();
```
msridhar, line 44: This `Set` of obligations should be an instance field in the `MustCallInvokedChecker`.  If we put it here, it will persist across the whole compilation, which is a potential memory leak of `Tree`s.  We should still keep the counter separately here.

msridhar (changes requested):

msridhar (approved):
looks great!

PR #311, opened 2021-02-12
Various fixes to ResetMustCall/constructors

kelloggm:
This PR covers several false positive causes in Zookeeper, all of which are related to `@ResetMustCall`:
* Reset must call required on constructor (2)
* Reassignment of owning field to null required reset must call (2)
* Required method not called on rhs of initializer of owning field (2)

It leaves in place some "expected" errors in the new tests that I don't think we actually need. These are errors about "re-assignments" to fields that occur in the constructor. I think such a "re-"assignment is safe if:
1. the field is defined in the class (i.e. the field isn't defined in a superclass)
2. the initializer is either not present or a `null` literal
3. the constructor doesn't dispatch to another constructor that might set the field (i.e. there is no `this(...)` call at the start of the constructor - calls to `super()` are okay because of 1.)
4. no other methods have yet been called in the constructor (since they might also assign the field - although we could probably weaken this to "no reset must call methods have been called" and be okay)

I haven't implemented these rules yet, but if you think they're sound then I could do so (at least the 2 false positives above in the "reset must call on constructor" category will remain if I don't).


msridhar (approved):
LGTM.  And I agree with the rules you propose to allow for setting a field in a constructor

kelloggm:
I'm going to merge this now and write the rules outlined above in a followup PR.

PR #312, opened 2021-02-12
no warning on MCC with this

kelloggm:
fixes one FP in Zookeeper

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -353,7 +353,7 @@ class MustCallInvokedChecker {
       MethodInvocationTree methodInvokeTree = (MethodInvocationTree) callTree;
       return TreeUtils.isSuperConstructorCall(methodInvokeTree)
           || TreeUtils.isThisConstructorCall(methodInvokeTree)
-          || returnTypeIsMustCallChoiceWithOwningFieldOrThis((MethodInvocationNode) node)
+          || returnTypeIsMustCallChoiceWithIgnorable((MethodInvocationNode) node)
           || hasNotOwningReturnType((MethodInvocationNode) node);
     }
     return false;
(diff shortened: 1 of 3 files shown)


msridhar (commented):
It seems that this change is 100% fine if `this` can _never_ be an `@Owning` pointer.  If `this` can be `@Owning` it seems like hypothetically this could lead to false positives, e.g., if one writes `this.getInputStream().close()` within the `MCCWithThis` test class.  Is my understanding correct?

kelloggm:
You're right, we'd issue a false positive in that case. I don't think that code will come up in practice, though, and if it does I'm fine with issuing a false positive.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -353,24 +353,28 @@ private boolean shouldSkipInvokeCheck(Node node) {
...
-   * the other must call choice is an owning field.
+   * the other must call choice is an owning field or this.
    *
    * @param node a method invocation node
-   * @return if this is the invocation of a method whose return type is MCC with an owning field
+   * @return if this is the invocation of a method whose return type is MCC with an owning field or
+   *     this
    */
-  private boolean returnTypeIsMustCallChoiceWithOwningField(MethodInvocationNode node) {
+  private boolean returnTypeIsMustCallChoiceWithOwningFieldOrThis(MethodInvocationNode node) {
```
msridhar, line 370: Could we rename this to `returnTypeIsMustCallChoiceWithOwningFieldOrNotOwning`?  And then indicate with a comment in the method body that for now, we assume that `this` is never owning?  No need to add any extra logic, just want to make the reasoning a bit clearer

msridhar (changes requested):
One hopefully minor ask

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
msridhar: We could also write that if `this` is owning it could lead to false positives

kelloggm: I actually renamed it to be more broad, because I realized that it doesn't matter if the field is owning - anything that's MustCallChoice with a field or `this` can be ignored.

If it's an owning field, then the class itself is responsible for closing it in its must-call method(s).

If it's a non-owning field, then this class can't be responsible - it must be a cache, or something, so we should ignore something that's MCC with that, too. I added a test case for this, too.

For `this`, I don't think it ever makes sense to have the `this` pointer be owning - that goes against the logic of must-call, which suggests that the client of a given class is responsible for calling some method on any instances of that class - the `this` pointer in a class should never be responsible for fulfilling the must-call obligations of the class.

msridhar (approved):
Eventually we might be able to broaden this, but it's fine for now.

PR #317, opened 2021-02-15
make reset must call repeatable

kelloggm:
This was a lot more invasive of a change than I'd expected, which is especially galling given that this will reduce the FP count by at most one on Zookeeper. Still, I think it's something we should have.

There is one visible behavior change: unparseable expressions that are the target of an RMC method are now also flagged as non-owning. I think that's fine, since we're issuing errors in those cases anyway. See e.g. the change in `BindChannel.java`.

### DIFF
--- a/object-construction-checker/tests/mustcall/ResetMustCallRepeat.java
+++ b/object-construction-checker/tests/mustcall/ResetMustCallRepeat.java
@@ -1,0 +1,72 @@
+// A simple test that @ResetMustCall is repeatable and works as intended.
+
+import org.checkerframework.checker.mustcall.qual.*;
+import org.checkerframework.checker.calledmethods.qual.*;
+import org.checkerframework.checker.objectconstruction.qual.*;
+
+@MustCall("a")
+class ResetMustCallRepeat {
+
+    @ResetMustCall("this")
+    @ResetMustCall("#1")
+    void reset(ResetMustCallRepeat r) { }
+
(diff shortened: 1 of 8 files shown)


object-construction-checker/tests/mustcall/ResetMustCallRepeat.java
```diff
@@ -0,0 +1,72 @@
...
+    void reset(ResetMustCallRepeat r) { }
+
+    void a() { }
+
+    static @MustCall({}) ResetMustCallRepeat makeNoMC() {
+        return null;
+    }
+
+    static void test1() {
+        // :: error: required.method.not.called
```
msridhar, line 21: One could argue that this error should be reported at the `reset` call, not here, since it is the `reset` call that introduces the obligation.  (Same comment for several other error locations in this file.)  But that is open to discussion and does not need to be fixed in this PR

msridhar (approved):
LGTM, though I admit I didn't read every line of the change carefully.  One comment for future discussion (not for this PR)

object-construction-checker/tests/mustcall/ResetMustCallRepeat.java
kelloggm: That's a fair point, though I agree it would be a separate change. I'm not too concerned at the moment, but we might want to make a change like that in the future, for usability.

PR #319, opened 2021-02-16
Feature flag for ResetMustCall

kelloggm:
This was a bit more convoluted than I would have liked, since we load stub files via the `@StubFiles` annotation. This approach works, but it's kinda gross ¯\_(ツ)_/¯

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/SocketAccumulationFrames.astub
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/SocketAccumulationFrames.astub
@@ -1,0 +1,41 @@
+// This version of the standard assumptions for sockets is intended for use without the -AnoAccumulationFrames
+// argument to the checker. It does include @MustCall({}) annotations for no-argument socket constructors. The
+// checker uses it automatically.
+
+package java.net;
+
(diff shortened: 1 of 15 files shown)


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/SocketAccumulationFrames.astub
```diff
@@ -0,0 +1,41 @@
+// This version of the standard assumptions for sockets is intended for use without the -AnoAccumulationFrames
```
msridhar, line 1: Isn't this one for use _with_ accumulation frames?  I.e., in cases where accumulation frames are not disabled?

object-construction-checker/tests/noaccumulationframes/ConnectingSockets.java
```diff
@@ -0,0 +1,58 @@
+// a set of test cases that demonstrate that errors are actually insued in appropriate
```
msridhar, line 1:
```suggestion
// a set of test cases that demonstrate that errors are actually issued in appropriate
```

msridhar (approved):
Two documentation nits

PR #321, opened 2021-02-16
Temporaries for MustCallChecker

Nargeshdb:
We made MustCallChecker aware of temporary variables in this PR.  To do that, we create temporaries and add them into the store in MCC. There are two failing tests after these changes that @kelloggm is working on them.


object-construction-checker/tests/socket/BindChannel.java
```diff
@@ -15,7 +15,7 @@ static void test(InetSocketAddress addr, boolean b) {
             // it's okay to keep issuing this error; see the second method in this class
             // for an example of how to rewrite the code to avoid the parse error.
             //
-            // :: error: required.method.not.called
+            // TODO we should report a warning here
```
kelloggm, line 18: This is okay for now, because we report the error below. I'm investigating this.

object-construction-checker/tests/socket/HdfsReport1.java
```diff
@@ -0,0 +1,11 @@
+// Based on a false positive in hdfs
```
kelloggm, line 1: Should we include this unrelated test in this PR? I would think it would go in a separate PR.

object-construction-checker/tests/socket/HdfsReport2.java
```diff
@@ -0,0 +1,36 @@
+// Based on a false positive in hdfs
```
kelloggm, line 1: Same comment as HdfsReport1.java

object-construction-checker/tests/socket/HdfsReport3.java
```diff
@@ -0,0 +1,38 @@
...
+    public void dumpTreeRecursively(PrintWriter out, StringBuilder prefix,
+                                    int snapshotId) {
+//        printing something
+    }
+
+    // StringBuilder doesn't implement closeable
+    final private StringBuilder sb = new StringBuilder();
+    final private Formatter formatter = new Formatter(sb);
+
+}
```
kelloggm, line 38: nit: no newline

kelloggm (approved):

object-construction-checker/tests/socket/ZookeeperReport6.java
```diff
@@ -1,13 +1,13 @@
...
 
-// @skip-test
-
 import java.nio.channels.SocketChannel;
 import java.io.IOException;
 
 class ZookeeperReport6 {
     SocketChannel createSock() throws IOException {
         SocketChannel sock;
+        // false positive
```
msridhar, line 9: why change skipping the test to a comment about a false positive?

object-construction-checker/tests/mustcall/CommonModuleCrash.java
```diff
@@ -1,11 +1,9 @@
 import java.net.*;
 
 class CommonModuleCrash {
-    // :: error: required.method.not.called
```
msridhar, line 4: why this change?

object-construction-checker/tests/mustcall/CommonModuleCrash.java
```diff
@@ -1,11 +1,9 @@
...
     Socket bar = new Socket();
     static void baz(Socket s) {  }
     static {
-        // :: error: required.method.not.called
```
msridhar, line 8: why this change?

msridhar (commented):
Is this pull request complete?  I see code for introducing temp vars in the MCC, but I don't see any code in the MCC that actually uses it.  Is the idea that we would land this and then add actual support in the MCC in a follow-up?

object-construction-checker/tests/mustcall/CommonModuleCrash.java
Nargeshdb: Because new Socket creates unbound socket which is MCC{}

object-construction-checker/tests/mustcall/CommonModuleCrash.java
kelloggm: These were false positives - these sockets are unconnected.

Nargeshdb: same reason

kelloggm:
@msridhar though there's not any code in the MCC that uses tempvars directly, this change already has significant impact. That's because the MustCallInvokedChecker looks up the type of various nodes in the Must Call Checker's store. With this change, it's able to look up the value of temporary variables - so, for example, it's able to reason that an expression that's MCC with an unconnected socket or `StringReader` is also `@MustCall({})`. Previously, to support resets, it was necessary to assume the worst about expressions that weren't in the MC store; that is, if we didn't know the type of an expression, we had to assume it was reset somewhere and therefore had its class' default type. This change makes that no longer necessary in a lot of cases, improving precision.

msridhar (approved):

kelloggm:
@msridhar @Nargeshdb I've fixed the false positive in `ZookeeperReport6.java`, and updated the explanation for why `BindChannel.java` issues the error where it does. This PR should now be ready to merge.

object-construction-checker/tests/socket/ZookeeperReport6.java
kelloggm: because I was fixing it, see my latest commit

PR #322, opened 2021-02-16
use compiler messaging interface to print counts instead of printf

kelloggm:
this is slightly an anti-pattern, since it doesn't use the localizable interface, but this is the only API we have that doesn't require a source location at which to print

running the tests also no longer prints the counts directly, since these messages are below the logging level used by the testing framework (because we use `-Anowarn` in our test runs)

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
@@ -1,5 +1,6 @@
 package org.checkerframework.checker.objectconstruction;
 
+import static javax.tools.Diagnostic.Kind.NOTE;
 import static org.checkerframework.checker.mustcall.MustCallChecker.NO_ACCUMULATION_FRAMES;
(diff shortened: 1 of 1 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionChecker.java
```diff
@@ -106,8 +107,9 @@ public void reportError(Object source, @CompilerMessageKey String messageKey, Ob
   @Override
   public void typeProcessingOver() {
     if (hasOption(COUNT_MUST_CALL)) {
-      System.out.printf("Found %d must call obligation(s).%n", numMustCall);
-      System.out.printf(
+      message(NOTE, "Found %d must call obligation(s).%n", numMustCall);
+      message(
+          NOTE,
           "Found %d must call obligation(s) that were handled correctly.%n",
```
msridhar, line 113: While we're here, can we clarify this message?
```suggestion
          "Successfully verified %d must call obligation(s).%n",
```

msridhar (approved):
LGTM with one small additional change

PR #324, opened 2021-02-17
make the count an error, so that maven doesn't eat it

kelloggm:
I'm getting deja vu, since this definitely happened last time I did this too and I just forgot :(


kelloggm:
and the reason I didn't make this change before becomes apparent: this makes all our tests fail

msridhar (approved):
Will the compilation fail because of these `ERROR`-level messages?

msridhar:
> and the reason I didn't make this change before becomes apparent: this makes all our tests fail

maybe we just need to extract `javac` commands to get these counts in cases where maven is used?

kelloggm:
@msridhar I think this latest version is the sweet spot. I've made it a warning, which won't fail compilation (or our tests), but will still show up in Maven's logs

msridhar (approved):
re-stamping for WARNING level

PR #327, opened 2021-02-17
enforce requirements on constructor return types

### LINKED ISSUE #326: If a class has an @Owning field, don't allow @MustCall({}) on any constructor
If we allow `@MustCall({})` on a constructor of a class with an `@Owning` field _f_, there is no guarantee anymore that _f_'s `@MustCall` obligation will be satisfied.  We should extend our checking to handle this case.

kelloggm:
fixes #326 

I had expected the CF to just handle this (via the Must Call Checker), so I was surprised when the test I wrote didn't fail. The reason was that I had disabled the relevant checking in BaseTypeVisitor for the Must Call Checker, because it was overly conservative - it assumed that every constructor return needed to be top! That's clearly inconsistent with Must Call (or any bottom-default typechecker, for that matter).

Actually, the constructor return just needs to be a super type of the declared lower bound for the class. This PR implements that check for Must Call.

IMO this change should probably be upstreamed into the CF's `BaseTypeVisitor`, probably after 25 Feb

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallVisitor.java
@@ -18,6 +18,7 @@ import org.checkerframework.common.basetype.BaseTypeChecker;
 import org.checkerframework.common.basetype.BaseTypeVisitor;
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
+import org.checkerframework.javacutil.ElementUtils;
 import org.checkerframework.javacutil.TreePathUtil;
 import org.checkerframework.javacutil.TreeUtils;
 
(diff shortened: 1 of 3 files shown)


msridhar:
Would this work for cases like `ByteArrayOutputStream`, where `@MustCall("close")` is inherited from the `Closeable` interface but we want the objects to be `@MustCall({})`?

kelloggm:
> Would this work for cases like ByteArrayOutputStream, where @MustCall("close") is inherited from the Closeable interface but we want the objects to be @MustCall({})?

These cases already issue errors. See e.g. tests/mustcall/ZookeeperByteBufferInputStream.java, which demonstrates that we issue an error both at the MustCall annotation on the class - because it's weaker than the super class - and on the constructor, because the receiver of the implicit `super` call (e.g. the `this` object) isn't MC close.

kelloggm:
For user-defined classes like ByteArrayOutputStream, I think it's appropriate that errors need to be suppressed.

msridhar (approved):

PR #332, opened 2021-02-20
fix RMC soundness bug

kelloggm:
This did change where one error was issued in `BindChannel.java`, which seems to be the test case that changes the most. I think that's fine. The reason is that there's a two-alias set in defs before `bind` is called containing both locals, but after there's only the one-alias set containing the second local (so the error is issued at the second local instead of the first).

I believe that this aligns the implementation with what you were describing for the paper, for the most part - there's still the split between the MCC and MCIC, but that's probably okay.

I did keep the owning logic in place, so reset.not.owning errors do still get issued. The basic premise of the owning logic is that there has to be something which takes ownership of the result. I think we only track must-alias for things in the def set, but if that's not the case this might still be unsound.

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
@@ -61,19 +61,38 @@ public class MustCallTransfer extends CFTransfer {
 
         if (result.containsTwoStores()) {
           CFStore thenStore = result.getThenStore();
-          thenStore.clearValue(targetExpr);
-          thenStore.insertValue(targetExpr, defaultType);
+          lubWithStoreValue(thenStore, targetExpr, defaultType);
+
           CFStore elseStore = result.getElseStore();
-          elseStore.clearValue(targetExpr);
-          elseStore.insertValue(targetExpr, defaultType);
+          lubWithStoreValue(elseStore, targetExpr, defaultType);
         } else {
           CFStore store = result.getRegularStore();
-          store.clearValue(targetExpr);
-          store.insertValue(targetExpr, defaultType);
+          lubWithStoreValue(store, targetExpr, defaultType);
         }
       }
     }
     return result;
+  }
+
+  /**
+   * Takes the LUB of the current value in the store for expr, if it exists, and defaultType.
+   * Inserts the result into the store as the new value for expr.
+   *
+   * @param store a CFStore
+   * @param expr an expression that might be in the store
+   * @param defaultType the default type of the expression's static type
+   */
+  private void lubWithStoreValue(CFStore store, JavaExpression expr, AnnotationMirror defaultType) {
+    CFValue value = store.getValue(expr);
+    CFValue defaultTypeAsValue = analysis.createSingleAnnotationValue(defaultType, expr.getType());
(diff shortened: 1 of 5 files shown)


kelloggm:
@msridhar I ran with this change on Zookeeper, and it produces four new false positives. Of these one looks like this:

```
            this.ss = ServerSocketChannel.open();
            ss.socket().setReuseAddress(true);
            LOG.info("binding to port {}", addr);
            ss.socket().bind(addr); // This call to bind is now rejected
            ss.configureBlocking(false); // <- note that this is a TP...

```

The call to `bind` is now rejected, because `ss.socket()` isn't an owning expression, and we no longer allow resolving an owning expression via a resource alias, which is what's going on here.

I think this is a false positive, but I'd appreciate your opinion.

The other three cases are very similar. There's a class that extends `ServerSocket`. It overwrites most of `ServerSocket`'s methods to look like this:

```
        /**                                                                                                             
         * See {@link Socket#connect(SocketAddress, int)}. Calling this method does not trigger mode detection.         
         */
        @Override
        @ResetMustCall("this")
        public void connect(SocketAddress endpoint, int timeout) throws IOException {
            getSocketAllowUnknownMode().connect(endpoint, timeout);
        }
```

We issue errors on the three overwrites of `ServerSocket`'s reset methods, because we can't check that `getSocketAllowUnknownMode`'s result is okay. Here's its definition:

```
        @NotOwning private Socket getSocketAllowUnknownMode() {
            if (isSecureSocket()) {
                return sslSocket;
            } else { // Note: mode is UNKNOWN or PLAINTEXT                                                              
                return prependableSocket;
            }
        }
```

(those are both owning fields of the class)

msridhar (approved):
@kelloggm you might still be looking to test cases but I'm ok with landing this as-is whenever

msridhar:
```java
            this.ss = ServerSocketChannel.open();
            ss.socket().setReuseAddress(true);
            LOG.info("binding to port {}", addr);
            ss.socket().bind(addr); // This call to bind is now rejected
            ss.configureBlocking(false); // <- note that this is a TP...
```

One thing that bothers me about this example is that I don't know how to rewrite the code so that it would type check.  My intuition is that maybe we can eventually extend the checker to verify this case.  My argument is that since `ss.socket()` is MCC with an `@Owning` field, we should treat this case like code that overwrites a non-final `@Owning` field.  I.e., if there is a `@ResetMustCall("this")` annotation on the enclosing method, it should pass our checker.  I haven't worked out the details, though, and it's not urgent.

kelloggm:
> One thing that bothers me about this example is that I don't know how to rewrite the code so that it would type check. My intuition is that maybe we can eventually extend the checker to verify this case. My argument is that since ss.socket() is MCC with an @Owning field, we should treat this case like code that overwrites a non-final @Owning field. I.e., if there is a @ResetMustCall("this") annotation on the enclosing method, it should pass our checker. I haven't worked out the details, though, and it's not urgent.

I think this is a fundamental design problem with `@ResetMustCall`. It abstracts away the details of what is actually being reset, and in this case we need to know that bind is going to overwrite the underlying socket specifically, as opposed to the more general "reset" to the Socket object. In theory these two are equivalent - since `Socket` only has one resource - but the abstraction has no way of being aware of that (right now).

PR #336, opened 2021-02-22
make countMustCall only count java* types

kelloggm:
I also removed the cache of things that we've already counted, because with it there I couldn't get the numbers on the Counter Test to be correct.


Nargeshdb (approved):
LGTM

kelloggm:
With this branch, I see 232 MC obligations on ZooKeeper, which seems about right.

PR #341, opened 2021-02-25
Anonymized version of the repo

kelloggm:
Never merge this, I just opened the PR to make it easy to look at.

This is the result of my effort to anonymize the main repo. The plan is:
* you all look over this and check to see if you can think of anything I missed
* I checkout this branch in a fresh copy of the repo
* in the fse-main-307 github account I make a new repo called "plumber"
* I copy all the files from the clean repo on this branch to that
* that's our anonymized tool

To produce this, I first fixed everything I could think of off the top of my head, and then I deleted a bunch of irrelevant stuff, and then I rewrote the README. The I searched the repo for any term on this list and deleted any reference to it:
* martin
* kellogg
* kelloggm
* manu
* sridharan
* msridar
* narges
* shadab
* nargeshdb
* mernst
* mike
* ernst
* uw
* washington
* ucr
* riverside


msridhar (approved):
LGTM (will not land of course)

kelloggm:
I realized that I can't remove quite as much as I'd like, because we need to be able to install the current snapshot version of the checker to the local Maven repo in order to run the case studies.

Manu and I discussed it, and we decided that we'll just pretend that we hard forked the OCC for the purposes of anonymization. This isn't quite as solid as I'd like - the maven coordinates of "plumber" are still `net.sridharan:object-construction-checker`-but we think that saying that we just copied it and only did things locally if anyone asks is a perfectly reasonable defense.

PR #357, opened 2021-03-18
Add stubs for GZip streams

mernst:


kelloggm (changes requested):
Stubs for these classes are already present in this file, immediately above the proposed changes. It looks to me as if these stubs are identical. If so, please close this PR; if not, please merge these with the existing stubs.

PR #359, opened 2021-03-20
Remove unnecessary deps

msridhar:
We don't need these dependencies anymore as the functionality is implemented within the Checker Framework.


msridhar:
This is obviously fine so I'm going to merge it without review

PR #363, opened 2021-03-22
Explain false positives due to implicit vs explicit bounds

mernst:
Making `@MustCall` the default for `IMPLICIT_UPPER_BOUND` definitely reduces annotation burden in my case studies (and might do so for other case studies).
I'm not sure whether making it the default for `IMPLICIT_LOWER_BOUND` is desirable.

### DIFF
--- a/must-call-checker/README.md
+++ b/must-call-checker/README.md
@@ -97,3 +97,44 @@ all `Socket` objects are `@MustCall({"close"})`. Without special-casing string c
 
     Socket sock = ...;
     String str = "this is the result of my socket's toString() function, which is invoked implicitly: " + sock;
+
+#### Type parameters with implicit vs explicit bounds sometimes cause (fixable) false positives
+
+The defaulting rules for `@MustCall` sometimes produce unexpected errors in code
+that uses type parameters with implicit upper bounds (i.e. without an `extends` clause).
+The errors can be fixed by adding an explicit bound. For example, consider the following
+example of a class with a type parameter and a field of an interface type that uses that
+type parameter (from [plume-lib/plume-util](https://github.com/plume-lib/plume-util)):
+
+```java
+    class MultiRandSelector<T> {
+        private Partition<T, T> eq;
+    }
+
+    interface Partition<ELEMENT extends @Nullable Object, CLASS extends @Nullable Object> {}
+```
+
+Running the Must Call Checker on this code produces two unexpected errors, at each use
+of `T` in `MultiRandSelector`:
+
+```
+must-call-checker/tests/mustcall/PlumeUtilRequiredAnnotations.java:19: error: [type.argument.type.incompatible] incompatible type argument for type parameter ELEMENT of Partitioner.
+        private Partitioner<T, T> eq;
+                            ^
+  found   : T extends @MustCallUnknown Object
+  required: @MustCall Object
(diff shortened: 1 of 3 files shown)


kelloggm:
I've investigated this a bit, and I think this change will make our benchmarks (Zookeeper at least, possibly the others) significantly worse not just in terms of annotation burden but also in terms of false positives. Consider the following test code, simplified from one of the failing tests:

```
    void test(List<@MustCall("close") Socket> list) {
        for (Socket s : list) {
            try {
                s.close();
            } catch (IOException i) { }
        }
    }
```

I showed you all this example in our call on Monday, and we kind of hand-waved it away because we thought there was an annotation we could write to make this code typecheck. However, that isn't the case. Consider the error message that's issued here:

```
object-construction-checker/tests/socket/EnhancedFor.java:12: error: [type.argument.type.incompatible] incompatible type argument for type parameter E of List.
    void test(List<@MustCall("close") Socket> list) {
                   ^
  found   : @MustCall("close") Socket
  required: @MustCall Object
```

This error message makes it clear that the problem here is that the problem is the type of `List`'s type parameter, which can only be bottom (i.e. `@MustCall({})`) with this change. That's because the implicit upper bound for an unannotated type variable is bottom with this change, so no other types are permitted. Any usage of an unannotated generic type like `List` with a class whose lowerbound is anything other than bottom (like `Socket` in this example) is therefore going to be rejected, regardless of the annotations at the use site. To fix this with an annotation, we'd need to annotate `List` explicitly to have an upperbound that's something other than bottom. 

IMO, it's not desirable to have to annotate every collection type as having a different parameter upperbound, even if it saves us some annotations elsewhere.

I tested the hypothesis that the test failures were caused by the implicit lower bound that was added in this PR, since Mike expressed less confidence in it. Removing it has no effect on the test failures; they're all caused by the implicit upper bound change. I think the implicit lower bound was already bottom - otherwise, `List<Object>` (where `Object` is implicitly `@MustCall({}) Object`) would be rejected by the current implementation.

msridhar:
@kelloggm are you able to explain the failures that Mike sees when he doesn't write `@MustCall` as a bound on generic type parameters?  That's the part I don't get.  Is there really no way to pick reasonable defaults for generic type parameters for this type system, so you only need to write an explicit `@MustCall` annotation when you are passing in values that have a non-empty `@MustCall`?

kelloggm:
@msridhar, you asked:
> are you able to explain the failures that Mike sees when he doesn't write @MustCall as a bound on generic type parameters

I've done some more experimentation, including writing a test case and some documentation (which I've pushed to this PR's branch). I think the problem isn't related to needing to write an `@MustCall` annotation at all; rather, the issue is that we default implicit and explicit upper bounds differently, and the cases in `plume-util` include a mix of implicit and explicit upper bounds. Can you take a look at the notes I wrote in the README about it, and let me know if that clears up your confusion?

msridhar:
@kelloggm this makes more sense, though perhaps @mernst should check the documentation change to see if it makes sense to him.  I have another (minor?) question.  You wrote:

>  we default implicit and explicit upper bounds differently

Who is "we" here?  The Checker Framework?  The Must Call Checker?  In any case, it seems like this issue is not as broadly impactful as I thought before; it only comes up when implicit and explicit upper bounds are getting mixed.  So, for now, I am ok with not changing defaulting in the Must Call Checker and living with the extra annotations in `plume-util`.

mernst (commented):
The current (comment only) changes look good to me.

kelloggm (approved):

PR #369, opened 2021-04-02
Fix false positive when creating a must-call alias with a field

msridhar:
See the new test.  Before, we would report a false positive warning when creating a must-call alias with a field.  Since we handle `@Owning` fields completely separately, there is no need to track the result of a call when its `@MustCallAlias` parameter is the result of a field read.

Eventually, it may be cleaner to find a way to ignore field reads in one place, rather than adding special checks in multiple places.  But that would take more thought.

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -361,7 +361,11 @@ class MustCallInvokedChecker {
                 .toSet();
         defs.remove(setContainingMustCallAliasParamLocal);
         defs.add(newSetContainingMustCallAliasParamLocal);
-      } else if (!(sameResource instanceof LocalVariableNode)) {
+      } else if (!(sameResource instanceof LocalVariableNode
+          || sameResource instanceof FieldAccessNode)) {
+        // we do not track the temp var for the call if the MustCallAlias parameter is a local (that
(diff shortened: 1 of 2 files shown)


object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -361,7 +361,11 @@ private void updateDefsWithTempVar(Set<ImmutableSet<LocalVarWithTree>> defs, Nod
                 .toSet();
         defs.remove(setContainingMustCallAliasParamLocal);
         defs.add(newSetContainingMustCallAliasParamLocal);
-      } else if (!(sameResource instanceof LocalVariableNode)) {
+      } else if (!(sameResource instanceof LocalVariableNode
+          || sameResource instanceof FieldAccessNode)) {
+        // we do not track the temp var for the call if the MustCallAlias parameter is a local (that
+        // case is handled above; the local must already be in the defs) or a field (since handling
+        // of @Owning fields is a completely separate check)
```
kelloggm, line 368: Since this comment specifically mentions owning fields, but the check on line 365 that you've added is checking for any field, it might be worth calling out here that a resource-alias of a non-owning field never needs to be tracked (because it's non-owning).

kelloggm (approved):
one minor comment, feel free to ignore it if you think it's too obvious to bother mentioning (but it did trip me up for a second when I was reading the new code)

PR #373, opened 2021-04-05
Define RequiresCalledMethods

Nargeshdb:


object-construction-checker/tests/mustcall/RequiresCalledMethodsTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+        @CreatesObligation("this")
+        void overwriteFooWithoutReleasing() {
+            // :: error: contracts.precondition.not.satisfied
+            overwriteFooCorrect();
+        }
+
+        void releaseThenOverwriteFoo() {
+            releaseFoo();
+            // :: error: reset.not.owning
```
msridhar, line 38: Why do we get this error here?

msridhar (commented):
Nice that this is so small!  One question

object-construction-checker/tests/mustcall/RequiresCalledMethodsTest.java
msridhar: nevermind i get it now

msridhar (approved):
LGTM now.  We can land here for now, but @Nargeshdb will also open a PR on the Checker Framework to add this alongside the Called Methods Checker, where it belong

PR #375, opened 2021-04-06
update to Checker Framework version 3.12.0

kelloggm:
Sorry that there are some extraneous changes in the Must Call Checker itself (i.e. adding docs, etc) - I copied the versions of those files from my PR against the CF proper, so they follow its javadoc guidelines.

I left intact a few uses of deprecated `getElementValueArray` calls, to avoid duplicating code. We probably should remove these, but it's a pain, because the code is shared between two different checkers and their type factories are taken as parameters (i.e. as `GenericAnnotatedTypeFactory`) depending on which checker uses the (static) method. Both need to get the same element from several shared annotations, so to fix this we'd need to create a subclass of GATF for them both to extend that has `ExecutableElement` fields for each of the annotation elements that these methods need to use.

### DIFF
--- a/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
+++ b/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
@@ -1,7 +1,5 @@
 package org.checkerframework.checker.mustcall;
 
-import static javax.lang.model.element.ElementKind.PARAMETER;
-
 import com.sun.source.tree.CompilationUnitTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IdentifierTree;
@@ -21,6 +19,7 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeKind;
+import org.checkerframework.checker.mustcall.qual.CreatesObligation;
 import org.checkerframework.checker.mustcall.qual.InheritableMustCall;
 import org.checkerframework.checker.mustcall.qual.MustCall;
 import org.checkerframework.checker.mustcall.qual.MustCallAlias;
@@ -30,7 +29,6 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.checker.objectconstruction.qual.Owning;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
-import org.checkerframework.common.value.ValueCheckerUtils;
 import org.checkerframework.dataflow.cfg.block.Block;
 import org.checkerframework.dataflow.cfg.node.LocalVariableNode;
 import org.checkerframework.dataflow.cfg.node.Node;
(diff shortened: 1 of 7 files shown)


must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
```diff
@@ -149,22 +164,28 @@ public void addComputedTypeAnnotations(Element elt, AnnotatedTypeMirror type) {
   }
 
   /**
-   * Creates a new @MustCall annotation that is identical to the input, but does not have "close".
+   * Creates a new MustCall annotation that is identical to the input, but does not have "close".
```
mernst, line 167: I personally like the "@" character.  It is a bit more explicit.  In code like the Index Checker that has internal representations with names similar to annotation names, it helps to distinguish between an annotation and the representation of an annotation.

All that said, either way is OK.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
```diff
@@ -260,6 +292,14 @@ public static String standardizeAndViewpointAdapt(
     return handleStringConcatenation(super.visitStringConcatenateAssignment(n, in));
   }
 
+  /**
+   * Create a new result for a string concatenation that forces their type to always be bottom.
```
mernst, line 296: "their" => "its"

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
```diff
@@ -289,6 +329,13 @@ public void updateStoreWithTempVar(TransferResult<CFValue, CFStore> result, Node
     }
   }
 
+  /**
+   * Either returns the temporary variable associated with node, or creates one if one does not
+   * exist. node must be an expression, not a statement.
```
mernst, line 334: I would put the documentation about `node` in the `@param` Javadoc tag.

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
```diff
@@ -289,6 +329,13 @@ public void updateStoreWithTempVar(TransferResult<CFValue, CFStore> result, Node
...
+   *
+   * @param node a node
+   * @return a temporary variable node representing node that can be placed into a store
```
mernst, line 337: This would be clearer if the second `node` were `{@code node}`.  As written, the two occurrences of "node" that mean different things are confusing.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
```diff
@@ -757,29 +769,36 @@ private void checkEnclosingMethodIsCreatesObligation(
...
-      String targetStr =
-          MustCallTransfer.standardizeAndViewpointAdapt(
-              targetStrWithoutAdaptation, currentPath, context);
+      String targetStr = null;
+      try {
+        targetStr =
+            StringToJavaExpression.atMethodBody(
+                    targetStrWithoutAdaptation, enclosingMethod, checker)
+                .toString();
+      } catch (JavaExpressionParseException e) {
```
mernst, line 799: Here and elsewhere, I'm not sure why the code suppresses an exception.  Doesn't the exception indicate a problem that either the user or the type system designer should know about?

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java
```diff
@@ -260,14 +262,13 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
-            if (oldFlowValues != null) {
-              // valuesAsList cannot have its length changed -- it is backed by an
-              // array.  getValueOfAnnotationWithStringArgument returns a new,
-              // modifiable list.
-              oldFlowValues.addAll(valuesAsList);
-              valuesAsList = oldFlowValues;
-            }
+                AnnotationUtils.getElementValueArray(
+                    anno, atypeFactory.calledMethodsValueElement, String.class);
+            // valuesAsList cannot have its length changed -- it is backed by an
```
mernst, line 267: Refill this comment.

mernst (approved):

must-call-checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
kelloggm: For this comment and the others in the Must Call Checker, I addressed them in https://github.com/typetools/checker-framework/pull/4484 but not here, since the copy of that checker in this repository will soon be deleted.

object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
kelloggm: We issue custom errors when `@CreatesObligation` annotation arguments are unparseable rather than propagating the exception. I found it more helpful, since the custom error messages can give better context.

PR #376, opened 2021-04-07
require that creates obligation targets aren't added when overriding methods

### LINKED ISSUE #368: Erroneous checking of `@CreatesObligation` on anonymous class
Test case:
```java
class COAnonymousClass {
  static class Foo {

    @CreatesObligation("this")
    void resetFoo() { }

    void other() {

      Runnable r = new Runnable() {
        @Override
        @CreatesObligation
        public void run() {
          resetFoo();
        }
      };
    }
  }
}
```
I checked this in on the `co-anonymous-class` branch.  The issue is that the code should not type check.  The call to `resetFoo()` in `run()` creates an obligation on `Foo.this`, not `this`.  I am not 100% sure what error we should report here.  We could run into the same issue with a non-static inner class or a lambda.

kelloggm:
This change will, I hope, fix #368. It doesn't address the problem directly, because I realized that the real issue behind #368 is that `run()` is overridden to add an `@CreatesObligation` annotation, but it could be dynamically-dispatched to by an object of static type `Runnable` but that is dynamically the anonymous class we defined. This problem isn't unique to anonymous classes; you'll see in the test cases that I was able to trigger the unsoundness without needing one.

I'm not convinced that my solution is great: I've restricted the places where `@CreatesObligation` can be written by adding a new kind of error that's issued when a method with fewer `@CreatesObligation` targets is overridden by one with more. Note that because `@CreatesObligation` is inheritable, in practice this means that it's only permitted to write a new obligation annotation on a method that isn't overriding anything. I think that's not too restrictive of a requirement, but I'm curious what you all think.

### DIFF
--- a/object-construction-checker/tests/mustcall/CreatesObligationInnerClass.java
+++ b/object-construction-checker/tests/mustcall/CreatesObligationInnerClass.java
@@ -2,7 +2,7 @@
 
 import org.checkerframework.checker.mustcall.qual.*;
 
-class COAnonymousClass {
+class CreatesObligationInnerClass {
     static class Foo {
 
         @CreatesObligation("this")
@@ -23,5 +23,24 @@ class COAnonymousClass {
 
         // If this call to run() were permitted with no errors, this would be unsound.
         void other2(Runnable r) { r.run(); }
+
+        /**
+         * non-static inner class
+         */
+        class Bar {
+
+            // this should be disallowed! not sure of the right error message
+            // :: error: creates.obligation.override.invalid
+            @CreatesObligation
+            void bar() {
+                resetFoo();
+            }
+        }
+
+        void callBar() {
+            Bar b = new Bar();
+            // If this call to bar() were permitted with no errors, this would be unsound.
+            b.bar();
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


msridhar (changes requested):
This is a good fix!  But it doesn't fully fix the original issue I identified, particularly in the case with non-static inner classes.  I added a new test to show the remaining issue.

kelloggm:
@msridhar I've fixed the underlying bug with `this` expressions now, so I think this is ready for review again. 

I also fixed up the code that handles `@CreateObligation` generally in a few places - before, it wasn't checking that every target was valid, and was instead issuing a `reset.not.owning` error only when ALL targets were invalid. That's fixed. 

I also simplified/fixed the code that handles `@CreateObligation` elsewhere, in particular the parsing and removing duplicate code for getting the expressions.

msridhar:
Will review later or tomorrow. Could you merge back the `always-call-checker` branch? There have been a couple changes landed there.

kelloggm:
@msridhar I've merged `always-call-checker` in. No hurries on the review - sorry this became a bit bloated. I can break it into 2 or 3 changes if it's too much.

object-construction-checker/tests/mustcall/CreatesObligationInnerClass.java
```diff
@@ -0,0 +1,29 @@
...
+    static class Foo {
+
+        @CreatesObligation("this")
+        void resetFoo() { }
+
+        /**
+         * non-static inner class
+         */
+        class Bar {
+            // this should be disallowed! not sure of the right error message
```
msridhar, line 15: we can get rid of this comment I think

msridhar (approved):
Looks great!  Just one tiny nit

PR #377, opened 2021-04-08
cleanup logic for handling assignments to local variables

msridhar:
The logic for handling assignments to local variables looks a lot more like what we have in the paper now (a gen-kill style), so I think it is cleaner.  As a small side bonus, we do not emit false positive warnings for assignments of the form `x = x` anymore.  But mainly I think the win is in the cleaner logic.  `SelfAssign.test1()` is based on real code from HBase and still needs to be handled, but I have left that for a follow-up.

I also took the opportunity to make the `tempVarToNode` map private and add accessor methods for it.


kelloggm (approved):
LGTM

Nargeshdb (approved):
LGTM. Thanks.

PR #378, opened 2021-04-10
Fixed handling of ternary expressions

msridhar:
This is a cleaner and more correct handling of ternary expressions in `MustCallInvokedChecker`.  In short, we now properly leverage the representation of ternary expressions in the Checker Framework CFG to handle each case separately.  See `SelfAssign.test1()` for a case we can handle now (based on real code in HBase) that we couldn't handle before.  This also re-uses the cleaned-up logic for assignments from #377.  I tried to add sufficient documentation to make the logic clear.


kelloggm (approved):
This looks great!

PR #389, opened 2021-04-27
When checking re-assignment to a field, get the called methods type from the store before evaluating the RHS

### LINKED ISSUE #381: False positive with `SocketFactory.createSocket()`
test case:
```java
  @MustCall("closeSocket")
  class SocketField {
    protected @Owning Socket socket = null;

    @CreatesObligation("this")
    protected void setupConnection(javax.net.SocketFactory socketFactory) throws IOException {
      this.socket.close();
      //this.socket = new Socket(); // no error, code verifies
      this.socket = socketFactory.createSocket(); // error on this line
    }
    @EnsuresCalledMethods(value = "this.socket", methods = "close")
    private void closeSocket() throws IOException {
      this.socket.close();
    }
  }
```
With this test we get a false positive `required.method.not.called` error at the assignment to `this.socket` inside `setupConnection()`.  If I comment that line and uncomment `this.socket = new Socket()` there is no error.  (The socket is never bound in this example, but I don't think that is relevant.)  Looking at `SocketAccumulationFrames.astub` I didn't see an obvious problem:

https://github.com/kelloggm/object-construction-checker/blob/always-call-checker/must-call-checker/src/main/java/org/checkerframework/checker/mustcall/SocketAccumulationFrames.astub

Not sure where to dig deeper.

kelloggm:
fixes #381. Previously, we got the called methods type from the store before the assignment node, which is after the RHS in the CFG (and therefore can have its store affected by side-effects from evaluating the RHS).

### DIFF
--- a/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
+++ b/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java
@@ -679,7 +679,6 @@ class MustCallInvokedChecker {
    *
    * @param node an assignment to a non-final, owning field
    * @param newDefs
-   * @param rhs
    */
   private void checkReassignmentToField(
       AssignmentNode node, Set<ImmutableSet<LocalVarWithTree>> newDefs) {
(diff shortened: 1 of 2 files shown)


msridhar (approved):
That's a nice clean fix!

PR #415, opened 2021-06-03
Missing scripts for artifact

kelloggm:
We didn't commit every script that we used to compute numbers from table 1, apparently. This corrects that mistake.

Also, I fixed `run-always-call-on-zookeeper.sh` to require that `JAVA8_HOME` is set, to prevent a problem that Mike encountered while double-checking my judgments (his machine had JAVA_HOME set to a Java 11 JDK, but zookeeper only builds with Java 8). @Nargeshdb and @msridhar, can you add similar logic to the scripts for hbase and hadoop, if they need a specific Java version?


msridhar (approved):
I added similar Java 8 changes for Hadoop and Hbase; @kelloggm you may want to check them.

We need to make sure this `JAVA8_HOME` env var is set correctly by default in our Docker image.

kelloggm:
@msridhar agreed, we need to make sure that the `JAVA8_HOME` variable is set. Using these scripts should make sure of that as long as we actually test things in the docker container :)

PR #418, opened 2021-06-04
add ensurescalledmethodsif on channel, selector, abstractselector isOpen method

kelloggm:


kelloggm:
These also need to go into typetools/jdk, but I'm leaving that for later

PR #419, opened 2021-06-04
Add failing test

mernst:
This indicates there is something wrong with https://github.com/kelloggm/object-construction-checker/pull/418.

@kelloggm Do you see what the problem is?  It isn't obvious to me offhand.

### DIFF
--- a/object-construction-checker/tests/mustcall/EnsuresCalledMethodsIfTest.java
+++ b/object-construction-checker/tests/mustcall/EnsuresCalledMethodsIfTest.java
@@ -1,0 +1,16 @@
+import java.nio.channels.SocketChannel;
+import org.checkerframework.checker.calledmethods.qual.EnsuresCalledMethods;
+
+class EnsuresCalledMethodsIfTest {
+
+    @EnsuresCalledMethods(value = "#1", methods = "close")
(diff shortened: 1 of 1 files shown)


kelloggm:
I don't know what's wrong here. I believe we've encountered this in the other case studies, too, and we don't know what the cause is. @msridhar does that sound right?

kelloggm:
If so, we should open a CF issue for this and reference that in the warning suppression comments.

msridhar:
> I don't know what's wrong here. I believe we've encountered this in the other case studies, too, and we don't know what the cause is. @msridhar does that sound right?

@Nargeshdb did see a case involving `isOpen()` in another case study, but I can't recall if we already tried adding `@EnsuresCalledMethodsIf` to fix it.  (Honestly I didn't even remember that we introduced `@EnsuresCalledMethodsIf`.)  

It is not obvious to me what the bug is here.  Yes, I think opening a CF issue at this point is the right thing to do.

kelloggm:
I've opened a CF issue: https://github.com/typetools/checker-framework/issues/4699

mernst:
Per my comment on https://github.com/typetools/checker-framework/issues/4699, I believe this is correct behavior.

PR #420, opened 2021-06-04
Add failing test

mernst:
Here is another problem with `@EnsuresCalledMethods`.

### DIFF
--- a/object-construction-checker/tests/mustcall/EnsuresCalledMethodsTest.java
+++ b/object-construction-checker/tests/mustcall/EnsuresCalledMethodsTest.java
@@ -1,0 +1,19 @@
+import java.io.IOException;
+import java.net.ServerSocket;
+import org.checkerframework.checker.calledmethods.qual.EnsuresCalledMethods;
+import org.checkerframework.checker.objectconstruction.qual.Owning;
+
+public class EnsuresCalledMethodsTest {
+
+    private @Owning ServerSocket ss;
+
+    @EnsuresCalledMethods(value = "ss", methods = "close")
+    public synchronized void stop() {
+        if (ss != null) {
+            try {
+                ss.close();
+            } catch (IOException e) {
+            }
+        }
+    }
+}


kelloggm:
This is a Checker Framework bug, not an OCC bug - there's nothing here that has anything to do with MustCall or the Resource Leak Checker. We should open an issue for this in the CF and then reference it in the SW justification.

msridhar:
I am surprised we don't have tests for this already.  I think this could be causing some false positives in our other case studies as well /cc @Nargeshdb

mernst:
When I remove the `@Owning` annotation and run on typetools (that is, Checker Framework master), the test passes.

kelloggm:
I suspect that the problem is that https://github.com/typetools/checker-framework/pull/4683 is not in the version of the CF that we're using (we're using 3.12.0, but that PR wasn't in a release until 3.14.0).

msridhar:
> I suspect that the problem is that [typetools/checker-framework#4683](https://github.com/typetools/checker-framework/pull/4683) is not in the version of the CF that we're using (we're using 3.12.0, but that PR wasn't in a release until 3.14.0).

Not sure this is the issue, as the logic in https://github.com/typetools/checker-framework/pull/4683 is already included in this checker, which subclasses the Called Methods Checker, see here:

https://github.com/kelloggm/object-construction-checker/blob/3c51a52ceba34ee7710cd74d181d2a5ec5ba9ca6/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/ObjectConstructionTransfer.java#L84

kelloggm:
@msridhar you're right, I'd forgotten. There must be some other difference between how this is handled here and how it's handled in the CF.

@mernst given our limited time budget, I think we should just live with these false positives for now. We can justify them with a link to this issue.

mernst:
We already have an appropriate false positive warning suppression for the relevant method in Zookeeper's `ObserverMaster.java`.

PR #421, opened 2021-06-04
fix bad script

kelloggm:


build.gradle
```diff
@@ -12,7 +12,7 @@ plugins {
 
 
 ext.versions = [
-    checkerFramework: "3.12.0",
+    checkerFramework: "3.14.0",
```
msridhar, line 15: we do not want this change at this point, right?

msridhar (changes requested):

build.gradle
kelloggm: no we do not

msridhar (approved):

PR #432, opened 2021-06-14
Add plume-util ablation script

kelloggm:


experimental-machinery/ablation/plume-util-ablation.sh
```diff
@@ -0,0 +1,47 @@
...
+    cat "${variant_name}-results" | grep "error:" | sort | uniq | wc -l
+
+    echo "old errors no longer produced by ${variant_name} configuration:"
+
+    grep "unneeded.suppression" "${variant_name}-results" | wc -l
+
+    echo "the result for the paper is the difference between these two numbers"
+
+}
+
```
Nargeshdb, line 38: I think we need this check here:

if [ "x${JAVA8_HOME}" = "x" ]; then
  echo "Please set JAVA8_HOME to run the checker on Plume Util. Plume Util requires Java 8."
  exit 1
fi

Nargeshdb (changes requested):

experimental-machinery/ablation/plume-util-ablation.sh
kelloggm: plume-util doesn't require Java 8. We only need that in the other benchmarks because they don't compile under Java 11, but plume-util should be fine on any recent Java version (I think).

Unless you tried running the script and it failed, but I doubt that because when I ran the experiment myself earlier I used Java 11.

Nargeshdb: So, do we need to check whether java version is 8 or 11?

kelloggm: I don't think a check is strictly necessary. If the current JDK isn't supported by the CF, then the CF gradle plugin (which plume-util uses to run our checker) will print a helpful error message, so I don't think we need to worry about it.

Nargeshdb (approved):
LGTM!

PR #471, opened 2021-08-23
update scripts and docker file

Nargeshdb:


fse-2021/LICENSE.md
```diff
@@ -1,5 +1,9 @@
-Our tool itself is distributed under an MIT license. This license also applies
-to our scripts, which were originally part of the tool repository.
+# Released under MIT License
```
kelloggm, line 1: I don't think this set of changes is correct.

I suggest retaining the original text, and then adding the license text underneath it. It's important to make clear that the case study programs are not released under MIT, which would be a violation of their license terms. (I also think the reviewer who complained about the license was being obtuse, but I don't object to including the actual license terms after the text that makes it clear to what it applies.)

fse-2021/LICENSE.md
```diff
@@ -1,5 +1,9 @@
...
 
-Our modifications to the case study programs (and those programs themselves)
-are distributed under their respective open-source licenses.
\ No newline at end of file
+Copyright (c) 2021 Martin Kellogg
```
kelloggm, line 3: All of our names should appear here, not just mine.

kelloggm (changes requested):

fse-2021/LICENSE.md
Nargeshdb: https://github.com/kelloggm/object-construction-checker/pull/471/commits/48c83052952914d857105be55e76926598875eb1

msridhar (approved):

msridhar:
LGTM, @kelloggm want to take another look?

kelloggm (approved):

